Issue #3, opened 2019-01-10
Implement or remove tvloader/config.go


swinslow:
Copied over from https://github.com/swinslow/spdx-go/issues/28:

In package `tvloader`, config.go has a Config structure that could be implemented to enable configuring the behavior of the tag-value loader. E.g., to flag it to stop on any error; to flag it to perform additional validation during the loading process; etc.

However, right now that file is unused. It should either be implemented, or else deleted.

swinslow:
Closed by #5

PR #5, opened 2019-03-03
Remove unused tvloader/config.go

abhishekspeer:
As of now, the tvloader/config.go script has nothing implemented, it can be removed for now.

Closes Issue #3


swinslow:
Works for me. Would like to implement a Config object for `tvloader` at some point, so that we can do things like configure whether the reader / parser should stop if it encounters an error. But for now, we're not using this, so good with removing it. Thanks!

Fixes #3

Issue #2, opened 2019-01-10
Builder uses local time rather than UTC for CreationInfo.Created


swinslow:
Copied over from https://github.com/swinslow/spdx-go/issues/7:

Currently, the `builder` package uses time.Now() to get the time for CreationInfo.Created. According to 2.9.1, it should be in UTC. Note that just calling UTC() as part of getting the time doesn't solve this, presumably because it needs more info about the user's local time zone.

swinslow:
Closed by #6

PR #6, opened 2019-03-04
Fix local time issue in CreationInfo.Created

abhishekspeer:
Builder package will now use UTC for Creationinfo.Created instead of local time. 

Fixes issue #2


swinslow:
Thanks @abhishekspeer! I think this works. Thanks for taking the initiative on this one  =)

swinslow:
Fixes #2

PR #16, opened 2019-08-17
Converter from Intermediate to Main data structure

abhishekspeer:
Signed-off-by: abhishekspeer.abhishek.abhishekgaur.gaur@gmail.com

Transfer script added to Converter from Intermediate to Main data structure.

### DIFF
--- a/v0/rdfloader/rdf2v1/Annotation.go
+++ b/v0/rdfloader/rdf2v1/Annotation.go
@@ -5,27 +5,28 @@ import (
 )
 
 type Annotation struct {
-	Annotator         ValueStr
-	AnnotationType    ValueStr
-	AnnotationDate    ValueStr
-	AnnotationComment ValueStr
+	Annotator                ValueStr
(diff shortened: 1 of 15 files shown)


rtgdk:
@abhishekspeer Have you changed the main function to return the main data structure instead of intermediate?
Also, add some example files to run the code.

abhishekspeer:
Thanks again @rtgdk! I've added the changes as well as referenced a seperate PR that adds example script to run the loader.

swinslow:
Overall looks good for the time being -- I'll go ahead and merge.

There are several ways that the rdfloader is still making assumptions about RDF files that are specific to the example file you were working with, which won't hold in the normal case. In particular the assumption that there will only be one Snippet is incorrect (and therefore where a pointer to one snippet is being returned by the parsing functions will not work as desired for RDF files with >1 snippets).

However, since we're keeping this within the rdfreader branch for now I'm fine with merging. Thank you @abhishekspeer!

PR #14, opened 2019-08-16
Add rdfSaver Files

abhishekspeer:
Signed-off-by: abhishekspeer <abhishek.abhishekgaur.gaur@gmail.com>

RDF writer support is added via this pull request.

PR #17, opened 2019-08-17
Add Example for rdfReader

abhishekspeer:
Signed-off-by: abhishekspeer.abhishek.abhishekgaur.gaur@gmail.com

Added example to test RDF reader.


abhishekspeer:
@swinslow and @rtgdk,  I've added example script to use rdf loader via this PR. Please suggest changes if any. :)

swinslow:
Looks good! Two comments:
- I see that this is missing the SPDX license statement at the top of the file, but I saw also that you added that in 058a618 (#14) so no need to add it in this current PR.
- I think this is demonstrating some problems with the parsing. E.g., when I run it against the example RDF file you included, I see that it is printing two package names but both are named `glibc`. It appears that one of them should be named `Saxon` based on the sample file.

For the second point, see https://github.com/spdx/tools-golang/blob/be473af59e3a83f900b547f14f038891ffe96dd8/examples/7-load-Rdf/examplerdf2v1.rdf#L763

This is something we should address but I am guessing it is a problem with the rdfloader code, not in this example file. So I will go ahead and merge this PR.

PR #29, opened 2020-05-10
Change data model to use IDs as keys

### LINKED ISSUE #24: Consider changing data model to IDs as keys
Currently, the data model for spdx/tools-golang stores everything in slices. For example:

- [Document2_1](https://github.com/spdx/tools-golang/blob/b68821f66a8d47c441e3da9ab059205e7e30b3f4/spdx/document.go#L8) contains slices of packages, other license sections, relationships, etc.
- [Package2_1](https://github.com/spdx/tools-golang/blob/b68821f66a8d47c441e3da9ab059205e7e30b3f4/spdx/package.go#L118) contains a slice of Files

Although this makes it easy to maintain the ordering of the applicable elements, it also makes lookup extremely difficult. Finding an element with a particular SPDX ID requires walking through the entire slice.

A better method might be for the internal data representation to store SPDX elements in maps, where the SPDXID (or LicenseRef ID) is the key.

If ordering is important to maintain, then an "order" field could also be maintained when reading.

### LINKED ISSUE #28: Should add spdxlib utility package
Follow-on to issue #24 

It would be helpful to have a package of utility functions that make it easier to work with SPDX files. In particular, now that packages etc. are stored using their SPDX identifiers as keys, it would be helpful to be able to quickly find the SPDX identifier for any DESCRIBES relationships for a document -- and/or a slice of just the single Package ID in the document, if there's just one. I'm sure there are other useful utility functions like this that would be helpful.

swinslow:
This refactors all of the spdx/tools-golang packages to use element IDs (and, where applicable, optional document IDs) as keys for documents, packages, files and snippets, rather than having them as slices.

It also changes relationships and annotations to use document/element IDs for references rather than arbitrary strings.

It _**does not**_ change Other Licensing Information sections to use LicenseRef- IDs as keys. Other Licensing Information sections are still stored in a slice in the Document rather than a map. That should be fixed, but it's a separate issue.

This PR also bundles in the first function for spdxlib, to help with identifying which packages are DESCRIBES / DESCRIBED_BY an SPDX document. This was necessary to get several of the examples working, since previously they just pulled from the index 0 package in a document.

Fixes #24 

Fixes #28

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 .vscode
 .DS_Store
 scratch/*
+*.swp
+
--- a/builder/build.go
+++ b/builder/build.go
@@ -64,7 +64,7 @@ func Build2_1(packageName string, dirRoot string, config *Config2_1) (*spdx.Docu
 
 	doc := &spdx.Document2_1{
 		CreationInfo:  ci,
-		Packages:      []*spdx.Package2_1{pkg},
+		Packages:      map[spdx.ElementID]*spdx.Package2_1{pkg.PackageSPDXIdentifier: pkg},
 		Relationships: []*spdx.Relationship2_1{rln},
 	}
 
--- a/builder/build_test.go
+++ b/builder/build_test.go
@@ -5,6 +5,8 @@ package builder
 import (
 	"fmt"
 	"testing"
+
+	"github.com/spdx/tools-golang/spdx"
 )
 
 // ===== Builder top-level Document test =====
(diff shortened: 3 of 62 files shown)


swinslow:
The failing test is just for a decrease in code coverage, which I don't particularly care about here (and would actually be fine with removing that check... if or when I figure out how)  :)

I'm going to wait a bit before merging this, because I want to be ready to update a couple of other tools that use spdx/tools-golang before I merge.

swinslow:
I haven't yet updated the tools I mentioned [above](https://github.com/spdx/tools-golang/pull/29#issuecomment-626374401). But I'm going to go ahead and merge this change as it is relevant to some other tooling I'm working on, and I haven't seen any requests to hold off.

PR #43, opened 2020-11-08
Fix tvloader storing incorrect doc ID

swinslow:
Fixes #38 

Signed-off-by: Steve Winslow <steve@swinslow.net>

Issue #41, opened 2020-08-22
Handle additional checksums permitted by 2.2


swinslow:
In SPDX 2.1 and earlier, the only hash algorithms that were permitted for e.g. File checksums, Package checksums, etc. were SHA1 (mandatory), SHA256 (optional) and MD5 (optional). Because of this, those fields were represented as three separate fields, one for each checksum type (see e.g. [here](https://github.com/spdx/tools-golang/blob/5a270f97f4106cf989761f9143e23132c1bffb3c/spdx/file.go#L20)).

In SPDX 2.2, several additional algorithms were permitted to be used (see [section 4.4 of the spec](https://spdx.github.io/spdx-spec/4-file-information/#44-file-checksum)). The tools-golang model doesn't currently allow these to be used.

Rather than adding more optional fields for each algorithm, I'm more inclined to switch to something where there is a `Checksum` type defined, with something like the following:

```go
type Checksum2_2 struct {
    Type ChecksumType
    Value string
}
```

and then using a series of consts / enums to represent the different algorithms:

```go
// should this all be ChecksumType2_2, etc. instead of just ChecksumType?
type ChecksumType int

const (
    CHECKSUM_TYPE_SHA1 = iota
    CHECKSUM_TYPE_SHA256
    // . . .
)
```

and then changing the various Checksum fields in Package, File, etc. to be a slice of `Checksum2_2`.

@RishabhBhatnagar I'd be interested in your thoughts on this approach!

RishabhBhatnagar:
Having merged #60, we can close this issue.

PR #44, opened 2020-11-08
Change ExternalDocumentReferences to structured data

### LINKED ISSUE #39: Ambiguity in ExternalDocumentRef of the spdx data model
[tools-golang/spdx/creation_info.go#L31](https://github.com/RishabhBhatnagar/tools-golang/blob/5a270f97f4106cf989761f9143e23132c1bffb3c/spdx/creation_info.go#L31) expects the ExternalDocumentRefs to be a list of strings.
But from the [spdx-spec](https://spdx.dev/wp-content/uploads/sites/41/2017/12/spdxversion2.1.pdf#h.h430e9ypa0j9) of externalDocumentRef for RDF file, it can be seen that externalDocumentRef is made up of two properties viz:
1. SPDXDocument &
2. checksum

Spdx data model must be modified to accept ExternalDocumentReferences as a List of struct having two described attributes and not a List of strings.

swinslow:
Fixes #39 

### DIFF
--- a/spdx/creation_info.go
+++ b/spdx/creation_info.go
@@ -28,7 +28,7 @@ type CreationInfo2_1 struct {
 
 	// 2.6: External Document References
 	// Cardinality: optional, one or many
-	ExternalDocumentReferences []string
+	ExternalDocumentReferences map[string]ExternalDocumentRef2_1
 
 	// 2.7: License List Version
 	// Cardinality: optional, one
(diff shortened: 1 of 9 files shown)


swinslow:
Hi @RishabhBhatnagar, take a look at this one too when you have a minute. This should resolve #39 by restructuring the ExternalDocumentReferences to be structured data instead of unparsed strings.

For now I'm just using a string to hold the algorithm type (as "Alg"). We may want to consider converting this into structured data itself, but that's being tracked in #41 and I think is a bit messier of a change so I didn't tackle it here.

Same comments as in #43 about the "error" from Coveralls test coverage decrease, I will look into adjusting that so that minor changes don't result in a failing report.

RishabhBhatnagar (approved):
Thanks a lot for creating a new type for ExternalDocumentReference!

Although, there is a repitition of the code for 2v1 and 2v2. We can atleast keep common utility functions (as the ExternalDocumentReferences don't differ across 2v1 and 2v2 if I recall correctly).

Having common utility functions like `extractExternalDocumentReference` will also enable rdfloader to re-use the component.

We will do it at some point later in time. For now, the changes are splendid. Thanks again!

PR #46, opened 2020-11-13
Gordf

RishabhBhatnagar:
The tool will be able to read and validate SPDX RDF/XML 2.2 documents using a multi-phase concurrent parser for rdf files [(gordf)](https://www.github.com/spdx/tools-golang) .


swinslow:
This is fantastic, thanks @RishabhBhatnagar! I know you've put an unbelievable about of work into getting this working. Let me take a closer look at it and I'll circle back shortly if I have any questions or comments, otherwise I'll go ahead and merge it.

swinslow:
@RishabhBhatnagar, this is great. I've taken a brief look through the PR, and I've also checked it out and ran some tests myself.

There are some issues we will need to address. In particular, testing it against the standard "example" file at https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXRdfExample-v2.2.spdx.rdf.xml, some data is not getting parsed: I am not seeing the Packages getting parsed that are defined in that RDF file. There are a handful of other things I'm seeing.

But those are all issues that we can address going forward. I think that it will be best to go ahead and merge this into the main branch now, and then we can file issues and make fixes there rather than maintaining a separate branch. I'll go ahead and merge now.

Just wanted to reiterate -- this is really great work, between everything in this PR and also writing [gordf](https://github.com/spdx/gordf) from scratch. Very well done!

PR #54, opened 2020-11-15
Fixes #51

### LINKED ISSUE #51: rdfloader: Missing Packages from parsing example file
When running rdfloader on the [canonical sample SPDX RDF file](https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXRdfExample-v2.2.spdx.rdf.xml), the parsing completes without any error message. However, not all of the data appears to be getting transferred into the main data model.

The following sections are getting filled in:
* Creation Info
* Unpackaged Files
* Other Licenses
* Relationships
* Annotations

However, there are no Packages in the resulting parsed Document2_2 object. The sample document does define a couple of Packages, so these should be showing up:

https://github.com/spdx/spdx-spec/blob/6b44f663463b6f351e264435baf9a5c3b302cd83/examples/SPDXRdfExample-v2.2.spdx.rdf.xml#L258

https://github.com/spdx/spdx-spec/blob/6b44f663463b6f351e264435baf9a5c3b302cd83/examples/SPDXRdfExample-v2.2.spdx.rdf.xml#L264

https://github.com/spdx/spdx-spec/blob/6b44f663463b6f351e264435baf9a5c3b302cd83/examples/SPDXRdfExample-v2.2.spdx.rdf.xml#L1223

It also looks like it is not parsing some Snippets or File references, for example:

https://github.com/spdx/spdx-spec/blob/6b44f663463b6f351e264435baf9a5c3b302cd83/examples/SPDXRdfExample-v2.2.spdx.rdf.xml#L8

https://github.com/spdx/spdx-spec/blob/6b44f663463b6f351e264435baf9a5c3b302cd83/examples/SPDXRdfExample-v2.2.spdx.rdf.xml#L15

I note that all of these appear outside the `<spdx:SpdxDocument>` tag defined at [line 990](https://github.com/spdx/spdx-spec/blob/6b44f663463b6f351e264435baf9a5c3b302cd83/examples/SPDXRdfExample-v2.2.spdx.rdf.xml#L990).

I don't know enough about RDF files to know if that's relevant, but I'm guessing it could be.

RishabhBhatnagar:


swinslow:
Thanks @RishabhBhatnagar! I just tried this out on the sample SPDX file from #51, and it definitely appears to fix _most_ of the missing data I mentioned there.

However, when translating it from RDF to tag value, I do note that some data is not present in the output. This could be an issue on the tag-value side, but from testing I _think_ it's an issue with the RDF reader. I'll include more detailed comments in the ongoing thread in #51.

Thanks again for this! I'll go ahead and merge it now.

PR #56, opened 2020-11-18
Enable Caching And Parse Cyclic Dependent Objects

### LINKED ISSUE #51: rdfloader: Missing Packages from parsing example file

### LINKED ISSUE #55: rdfloader: Issue With The Licenses
Licenses, when associated with itself, will cause infinite recursion in the tool.

Licenses are implemented without using pointers. Hence, everytime a new License object is created.
It is not possible to cache it. As, everytime a new object will be created and adding data to it won't affect the old one causing inconsistency of data.

@swinslow , Is it possible for any spdx document to have cyclic dependency on licenses?
Only way I can see that happening is via a `ConjunctiveLicenseSet` or `DisjunctiveLicenseSet`.

@swinslow, should I assume that there might be a case where a license is associated with itself in some way?

RishabhBhatnagar:
- File, Package, Relationship can have cyclic dependent objects.
 - Cyclic dependent License will raise an error.
 - Double computation for File, Package, Relationship, and License is enabled.

This fixes #55 and attempts to fix #51.

@swinslow, this time, I've personally checked if the parsed document has all the bits from the [canonical sample SPDX RDF file](https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXRdfExample-v2.2.spdx.rdf.xml).
Please let me know if there are still any missing elements in the parsed document.

Also, please feel free to try it out using other documents as well. 

Signed-off-by: Rishabh Bhatnagar <bhatnagarrishabh4@gmail.com>

### DIFF
--- a/rdfloader/parser2v2/parse_file.go
+++ b/rdfloader/parser2v2/parse_file.go
@@ -13,6 +13,25 @@ import (
 func (parser *rdfParser2_2) getFileFromNode(fileNode *gordfParser.Node) (file *spdx.File2_2, err error) {
 	file = &spdx.File2_2{}
 
+	currState := parser.cache[fileNode.ID]
+	if currState == nil {
+		// this is the first time we are seeing this node.
+		parser.cache[fileNode.ID] = &nodeState{
+			object: file,
+			Color:  WHITE,
+		}
+	} else if currState.Color == GREY {
+		// we have already started parsing this file node and we needn't parse it again.
+		return currState.object.(*spdx.File2_2), nil
+	}
+
+	// setting color to grey to indicate that we've started parsing this node.
+	parser.cache[fileNode.ID].Color = GREY;
+
+	// setting color to black just before function returns to the caller to
+	// indicate that parsing current node is complete.
+	defer func() { parser.cache[fileNode.ID].Color = BLACK }()
+
 	err = setFileIdentifier(fileNode.ID, file) // 4.2
 	if err != nil {
 		return nil, err
--- a/rdfloader/parser2v2/parse_file_test.go
+++ b/rdfloader/parser2v2/parse_file_test.go
@@ -494,6 +494,25 @@ func Test_rdfParser2_2_getFileFromNode(t *testing.T) {
 		t.Errorf("expected %s, found %s", expectedLicenseInfoInFile, file.LicenseInfoInFile[0])
 	}
 
+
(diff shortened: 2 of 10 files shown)


swinslow:
Thanks @RishabhBhatnagar! From what I can tell, this looks great and yes, it appears to pick up everything that was missing previously. Great job!

I have a couple of other comments, but those can be going-forward fixes and I'll file separate issues where we can discuss those.

PR #60, opened 2021-02-28
[Optional Upgrade] Checksum Struct done .

### LINKED ISSUE #49: [Optional Upgrade] Checksum Struct
Checksums are currently stored in coalesced strings. They can be upgraded to use a struct.

Following seems like a suitable implementation:
```go
type Checksum {
    Algorithm, Value String
}
```
For now, the Algorithm field is kept as a String.

We can also make it a field that allows only algorithms defined by the model in an enum.
Implementation goes as follows:
```golang
type ChecksumAlgorithm int
const (
    MD5 ChecksumAlgorithm = iota
    SHA1
    SHA512
    MD2
    // other algorithms...
)

type Checksum {
    Algorithm ChecksumAlgorithm
    Value String
}
```

specter25:
this fixes #49

### DIFF
--- a/spdx/checksum.go
+++ b/spdx/checksum.go
@@ -3,24 +3,24 @@
 package spdx
 
 // ChecksumAlgorithm2_2 represents the algorithm used to generate the file checksum in the Checksum2_2 struct.
-type ChecksumAlgorithm2_2 string
+type ChecksumAlgorithm string
 
 // The checksum algorithms mentioned in the spdxv2.2.0 https://spdx.github.io/spdx-spec/4-file-information/#44-file-checksum
 const (
-	SHA224 ChecksumAlgorithm2_2 = "SHA224"
-	SHA1                        = "SHA1"
-	SHA256                      = "SHA256"
-	SHA384                      = "SHA384"
-	SHA512                      = "SHA512"
-	MD2                         = "MD2"
-	MD4                         = "MD4"
-	MD5                         = "MD5"
-	MD6                         = "MD6"
+	SHA224 ChecksumAlgorithm = "SHA224"
+	SHA1                     = "SHA1"
+	SHA256                   = "SHA256"
+	SHA384                   = "SHA384"
+	SHA512                   = "SHA512"
+	MD2                      = "MD2"
+	MD4                      = "MD4"
+	MD5                      = "MD5"
+	MD6                      = "MD6"
 )
 
 //Checksum2_2 struct Provide a unique identifier to match analysis information on each specific file in a package.
 // The Algorithm field describes the ChecksumAlgorithm2_2 used and the Value represents the file checksum
-type Checksum2_2 struct {
-	Algorithm ChecksumAlgorithm2_2
+type Checksum struct {
+	Algorithm ChecksumAlgorithm
 	Value     string
 }
--- a/spdx/checksum.go
+++ b/spdx/checksum.go
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+
+package spdx
+
+// ChecksumAlgorithm2_2 represents the algorithm used to generate the file checksum in the Checksum2_2 struct.
+type ChecksumAlgorithm2_2 string
+
+// The checksum algorithms mentioned in the spdxv2.2.0 https://spdx.github.io/spdx-spec/4-file-information/#44-file-checksum
+const (
+	SHA224 ChecksumAlgorithm2_2 = "SHA224"
+	SHA1                        = "SHA1"
+	SHA256                      = "SHA256"
+	SHA384                      = "SHA384"
+	SHA512                      = "SHA512"
+	MD2                         = "MD2"
+	MD4                         = "MD4"
+	MD5                         = "MD5"
+	MD6                         = "MD6"
+)
+
+//Checksum2_2 struct Provide a unique identifier to match analysis information on each specific file in a package.
+// The Algorithm field describes the ChecksumAlgorithm2_2 used and the Value represents the file checksum
+type Checksum2_2 struct {
+	Algorithm ChecksumAlgorithm2_2
+	Value     string
+}
--- a/spdx/file.go
+++ b/spdx/file.go
@@ -105,9 +105,11 @@ type File2_2 struct {
 
 	// 4.4: File Checksum: may have keys for SHA1, SHA256 and/or MD5
 	// Cardinality: mandatory, one SHA1, others may be optionally provided
-	FileChecksumSHA1   string
-	FileChecksumSHA256 string
-	FileChecksumMD5    string
+	// FileChecksumSHA1   string
+	// FileChecksumSHA256 string
+	// FileChecksumMD5    string
+
+	FileChecksums map[ChecksumAlgorithm2_2]Checksum2_2
 
 	// 4.5: Concluded License: SPDX License Expression, "NONE" or "NOASSERTION"
 	// Cardinality: mandatory, one
--- a/spdx/file.go
+++ b/spdx/file.go
@@ -105,11 +105,8 @@ type File2_2 struct {
 
 	// 4.4: File Checksum: may have keys for SHA1, SHA256 and/or MD5
 	// Cardinality: mandatory, one SHA1, others may be optionally provided
-	// FileChecksumSHA1   string
-	// FileChecksumSHA256 string
-	// FileChecksumMD5    string
 
-	FileChecksums map[ChecksumAlgorithm2_2]Checksum2_2
+	FileChecksums map[ChecksumAlgorithm]Checksum
 
 	// 4.5: Concluded License: SPDX License Expression, "NONE" or "NOASSERTION"
 	// Cardinality: mandatory, one
--- a/spdx/file.go
+++ b/spdx/file.go
@@ -105,7 +105,6 @@ type File2_2 struct {
 
 	// 4.4: File Checksum: may have keys for SHA1, SHA256 and/or MD5
 	// Cardinality: mandatory, one SHA1, others may be optionally provided
-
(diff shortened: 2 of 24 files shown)


specter25:
@RishabhBhatnagar , @swinslow can you please review this pr , and suggest me possible improvements .

specter25:
@RishabhBhatnagar , @swinslow i have fixed the failing tests .

RishabhBhatnagar:
I'm not sure why is coveralls is showing a decreased coverage due to a snippet that was not modified in this commit.
In order for the build case to pass, you can try doing either of the following:
1. If possible, write a test case covering the uncovered statements (3  lines).
2. squashing the last two commits might help.

specter25:
@RishabhBhatnagar sir i have written a the test for that particular situation . Can you please review the pull request once and suggest any other improvement to be done  .

specter25:
@RishabhBhatnagar   Sir if this pr is merged i can work on #41 as well .

spdx/checksum.go
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+
+package spdx
+
+type ChecksumAlgorithm2_2 int
```
RishabhBhatnagar, line 5: If a checksumAlgorithm type is set as an int, we loose the ability to distinguish the algorithm type in runtime unless we compare the algorithm type against all the possible algorithms.

The point I'm trying to put across is that, we would be identifying each ChecksumAlgorithm2_2 as an int. If we were to extent this model to let the ChecksumAlgorithm2_2 object decide it's name, it won't be inherently possible. This also would break the implicit typecasting.

For example: 
```go
fmt.Println(Checksum2_2{
	Algorithm: SHA1,
	Value:     "",
})
```

This would print `{0 }`

To resolve this, there can either be two approaches:
1. Add another function of ChecksumAlgorithm2_2 that maps the algorithm type to their string equivalent.
2. Let the enum type be a string instead of int.

spdx/file.go
```diff
@@ -105,9 +105,11 @@ type File2_2 struct {
...
 	// 4.4: File Checksum: may have keys for SHA1, SHA256 and/or MD5
 	// Cardinality: mandatory, one SHA1, others may be optionally provided
-	FileChecksumSHA1   string
-	FileChecksumSHA256 string
-	FileChecksumMD5    string
+	// FileChecksumSHA1   string
+	// FileChecksumSHA256 string
+	// FileChecksumMD5    string
+
+	FileChecksums []Checksum2_2
```
RishabhBhatnagar, line 112: Since each file cannot be associated with more than one checksums of same algorithm, it would be better to implement FileChecksums as  a map of ChecksumAlgorithm2_2 and Checksum2_2.

That would enable us to quickly get checksum for an algorithm. For example: if I want to access MD5 checksum value, we can do `fileObj.FileChecksum[MD5].Value` to get the value.

utils/verification.go
```diff
@@ -53,7 +53,11 @@ func GetVerificationCode2_2(files map[spdx.ElementID]*spdx.File2_2, excludeFile
 			return "", fmt.Errorf("got nil file for identifier %v", i)
 		}
 		if f.FileName != excludeFile {
-			shas = append(shas, f.FileChecksumSHA1)
+			for _, checksum := range f.FileChecksums {
```
RishabhBhatnagar, line 56: This for loop could've been avoided if the `FileChecksums` was a map.

utils/verification_test.go
```diff
@@ -120,27 +120,57 @@ func TestPackage2_2CanGetVerificationCode(t *testing.T) {
 		"File0": &spdx.File2_2{
 			FileName:           "file2.txt",
 			FileSPDXIdentifier: "File0",
-			FileChecksumSHA1:   "aaaaaaaaaabbbbbbbbbbccccccccccdddddddddd",
+			// FileChecksumSHA1:   "aaaaaaaaaabbbbbbbbbbccccccccccdddddddddd",
+			FileChecksums: []spdx.Checksum2_2{
+				spdx.Checksum2_2{
+					Algorithm: 1,
```
RishabhBhatnagar: "Algorithm" field should be populated using the defined constants (that is `SHA1`).
SHA1 in the current implementation is iota whose value is 0.

RishabhBhatnagar (changes requested):

RishabhBhatnagar:
@swinslow, do we want separate checksum structures for spdx versions 2.2 and 2.1? In my opinion, one checksum struct should be enough.

specter25:
okay sir on it . Will do the fixes asap .

specter25:
@RishabhBhatnagar sir i have done the requested changes , you can review them .

RishabhBhatnagar (approved):

RishabhBhatnagar (changes requested):
Thank You for your contributions and prompt response.

To maintain consistency across the project, we try to follow some of the conventions (mentioned [here](https://chris.beams.io/posts/git-commit/)). It would really help out everyone if you can make the necessary changes to follow them.

spdx/file.go
```diff
@@ -104,12 +104,12 @@ type File2_2 struct {
 	FileType []string
 
 	// 4.4: File Checksum: may have keys for SHA1, SHA256 and/or MD5
-	// Cardinality: mandatory, one SHA1, others may be optionally provided
+	// Cardinalitsy: mandatory, one SHA1, others may be optionally provided
```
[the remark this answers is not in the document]
specter25: really sorry for this typo

specter25:
- @RishabhBhatnagar i have made the changes , and tried to fix the commit messages as well. Hope it works .

specter25:
@RishabhBhatnagar sir i have made the necessary changes . I have tried to fix the git commit history as well , but not sure about the force push . Hope it works .

swinslow:
Hi @specter25 and @RishabhBhatnagar, thank you both for all your work on this. I'm (finally!) getting a chance to take a look back at this now.

I'll review more closely and will circle back with some thoughts, though I did have one initial reaction to a question above from @RishabhBhatnagar:

> @swinslow, do we want separate checksum structures for spdx versions 2.2 and 2.1? In my opinion, one checksum struct should be enough.

I agree -- I think one checksum struct is sufficient. Different versions of the SPDX spec may permit different checksum types depending on the situation, but I think it's fine here for us to have a single checksum struct type that is used for all of them.

specter25:
Sure @swinslow sir . Definitely up for suggestions .

swinslow:
@specter25, good work on pulling this together. Overall this looks great. I have two main comments for adjustments:

1. mentioned above -- yes, let's change `ChecksumAlgorithm2_2` to just `ChecksumAlgorithm` and `Checksum2_2` to just `Checksum`. That way, they should be more reusable for other versions of SPDX (such as 2.1).
2. I see that in many places you commented out the lines that were being replaced. Can you remove those lines, instead of commenting them out?

Those are the two things I noticed -- @RishabhBhatnagar may have additional comments.

Also, just as a heads-up, I've adjusted the test coverage check so that minor changes in coverage shouldn't trigger a failure any longer. That should make it easier so we aren't all running into issues trying to keep the coverage high with minor changes. :)

Thanks again for all your help!

specter25:
@swinslow sure sir , I will do the mentioned fixes as soon as possible . Thanks for reviewing the pr :)

specter25:
@swinslow , @RishabhBhatnagar  sir i have done the mentioned changes , you can give it a look :)

tvsaver/saver2v2/save_file_test.go
```diff
@@ -18,10 +18,22 @@ func TestSaver2_2FileSavesText(t *testing.T) {
...
+			spdx.SHA256: spdx.Checksum{
+				Algorithm: spdx.SHA256,
+				Value:     "11b6d3ee554eedf79299905a98f9b9a04e498210b59f15094c916c91d150efcd",
+			},
+			spdx.MD5: spdx.Checksum{
+				Algorithm: spdx.MD5,
+				Value:     "624c1abb3664f4b35547e7c73864ad24",
+			},
+		},
+
```
RishabhBhatnagar, line 35: Redundant NL.

tvsaver/saver2v2/save_file_test.go
```diff
@@ -218,8 +242,14 @@ func TestSaver2_2FileWrapsCopyrightMultiLine(t *testing.T) {
...
 		FileSPDXIdentifier: spdx.ElementID("File123"),
-		FileChecksumSHA1:   "85ed0817af83a24ad8da68c2b5094de69833983c",
-		LicenseConcluded:   "Apache-2.0",
+		FileChecksums: map[spdx.ChecksumAlgorithm]spdx.Checksum{
+			spdx.SHA1: spdx.Checksum{
+				Algorithm: spdx.SHA1,
+				Value:     "85ed0817af83a24ad8da68c2b5094de69833983c",
+			},
+		},
+
```
RishabhBhatnagar, line 251: Redundant NL again. Is this due to the IDE you're using?
We have similar NLs all over the commit.

RishabhBhatnagar (changes requested):

tvsaver/saver2v2/save_file_test.go
RishabhBhatnagar: @specter25 , you've done a great job. Appreciate your persistence.

specter25:
@RishabhBhatnagar sir sorry for the inconsistency in the code . I have tried to remove all the redundant NL . I have also combined the switch cases by the mechanism you suggested  :)

RishabhBhatnagar (approved):

swinslow (approved):

swinslow:
Looks good to me -- thank you @specter25 for all your help! Merging now.

specter25:
It was really a learning experience sir . Happy to contribute . Thanks @RishabhBhatnagar sir and @swinslow sir  for your help :)

PR #63, opened 2021-03-20
Fix special IDs for right-side 2.2 Relationships

### LINKED ISSUE #59: Relationship field in tag/value parser rejects valid values
Thanks for making this Go library! Having the ability to use SPDX files in Go is excellent.

I think I've found a bug in how the tag/value parser handles Relationship fields: [SPDX 2.2 specifies](https://spdx.github.io/spdx-spec/7-relationships-between-SPDX-elements/) this example is valid: 

```
Relationship: SPDXRef-BobBrowser CONTAINS NOASSERTION
```

However, with a file containing 

```
Relationship: SPDXRef-ion-tester CONTAINS NOASSERTION
```

the `tvloader.Load2_2(r)` call returns an error stating `missing SPDXRef- prefix for element identifier`.

The same file successfully passes with the [Java validator](https://github.com/spdx/tools-java):

```
$ java -jar /tmp/tools-java-0.0.6-jar-with-dependencies.jar Verify ./spdx-test-1.spdx TAG

This SPDX Document is valid.
```

I've made a reproduction repository at https://github.com/matthewkmayer/spdx-go-repro . It can be cloned then run with `go run .` to see the error.

I tried to find tests for this scenario in this repo but didn't find any. Perhaps guidance on how to find the relevant tests could help us solve this. 👍

swinslow:
In SPDX 2.2, the right-hand side of Relationships are not limited
to SPDX IDs; they can also include the special values NONE and
NOASSERTION.

To handle these, since Golang doesn't (to my knowledge) have a
concept of union types, and since I don't want to use interface{},
this commit instead adds a new SpecialID field to DocElementID.
When SpecialID is non-empty, it should be treated as being a
"special" ID value, and DocumentRefID / ElementRefID should be
ignored.

(Unfortunately, we can't just use ElementRefID == "NONE", etc.
for this purpose, because in theory an SPDX document could define
the identifier SPDXRef-NONE to mean something. Even though they
really, really shouldn't do that.)

This commit updates tvloader and tvsaver to appropriately handle
the possibility of NONE and NOASSERTION for this field.

Fixes #59 

### DIFF
--- a/spdx/identifier.go
+++ b/spdx/identifier.go
@@ -18,9 +18,14 @@ type ElementID string
 // present document.
 // DocElementIDs should NOT contain the mandatory 'DocumentRef-' or
 // 'SPDXRef-' portions.
+// SpecialID is used ONLY if the DocElementID matches a defined set of
+// permitted special values for a particular field, e.g. "NONE" or
+// "NOASSERTION" for the right-hand side of Relationships. If SpecialID
+// is set, DocumentRefID and ElementRefID should be empty (and vice versa).
 type DocElementID struct {
 	DocumentRefID string
 	ElementRefID  ElementID
+	SpecialID     string
 }
 
 // TODO: add equivalents for LicenseRef- identifiers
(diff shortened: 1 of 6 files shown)


swinslow:
Hi @matthewkmayer, this should address the issue you raised in #59. Grateful if you can take a look and let me know your thoughts. When I try it on my local machine, it can now successfully parse the file from the demo you shared in https://github.com/matthewkmayer/spdx-go-repro.

This PR is to merge into a new v0.1.1 branch (since I tagged and released v0.1.0 earlier today). The main branch has some other more significant API changes that are likely to get merged shortly, but I wanted to get this bug fix in for the existing API since you flagged it.

matthewkmayer:
Thanks! This change fixes #59 in [my test project](https://github.com/matthewkmayer/spdx-go-repro/tree/upgrade-test).

The addition of tests here make it easier for me to follow along with the changes as well. 👍

swinslow:
Ugh -- so sorry for the delay in merging this  :(  April was a long month.

@matthewkmayer thank you for responding so quickly on this. I'm merging it now into the 0.1.1 branch.

PR #64, opened 2021-03-28
Throw error for pkgs / files without SPDX IDs

### LINKED ISSUE #26: Should warn or error for pkgs / files without SPDX IDs
Placeholder for now, to handle after #24 is implemented:

Should be sure that tvparser either warns or errors out upon seeing a Package or File section that is incomplete due to lack of an SPDX ID.

For example, a new package section will be started when the next `PackageName` tag is seen, but if it goes on to the next Package or File without seeing an `SPDXID` tag for that first Package, then it will not be added to the Packages map for that document. The same applies for Files (but not relevant for Snippets, see https://github.com/spdx/tools-golang/issues/24#issuecomment-626216095)

Lack of an SPDX ID makes the document invalid, but consider whether the parser should treat it as an error, a warning, and/or configurable for either of those or to optionally ignore.

specter25:
This fixes #26

### DIFF
--- a/tvloader/parser2v1/parse_file_test.go
+++ b/tvloader/parser2v1/parse_file_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/spdx/tools-golang/spdx"
+	"github.com/spdx/tools-golang/tvloader/reader"
 )
 
 // ===== Parser file section state change tests =====
@@ -885,4 +886,60 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
 	}
 }
 
+func TestParser2_1FilesWithoutSpdxIdThrowError(t *testing.T) {
+	//case 1
+	// Last unpackaged file no packages in doc
+	// Last file of last package in the doc
+	var tvPairs []reader.TagValuePair
+	tvPair1 := reader.TagValuePair{Tag: "SPDXVersion", Value: "SPDX-2.1"}
+	tvPairs = append(tvPairs, tvPair1)
+	tvPair2 := reader.TagValuePair{Tag: "DataLicense", Value: "CC0-1.0"}
+	tvPairs = append(tvPairs, tvPair2)
+	tvPair3 := reader.TagValuePair{Tag: "SPDXID", Value: "SPDXRef-DOCUMENT"}
+	tvPairs = append(tvPairs, tvPair3)
+	tvPair4 := reader.TagValuePair{Tag: "FileName", Value: "f1"}
+	tvPairs = append(tvPairs, tvPair4)
+	_, err := ParseTagValues(tvPairs)
+	if err == nil {
+		t.Errorf("files withoutSpdx Identifiers getting accepted")
+	}
 
+	//case 2 : The previous file (packaged or unpackaged does not contain spdxID)
+	tvPair5 := reader.TagValuePair{Tag: "FileName", Value: "f2"}
+	tvPairs = append(tvPairs, tvPair5)
+	_, err = ParseTagValues(tvPairs)
+	if err == nil {
+		t.Errorf("%s", err)
+	}
+
+	//case 3 : Invalid file with snippet
+	//Last unpackaged file before the packges start
+	//Last file of a package and New package starts
+	sid1 := spdx.ElementID("s1")
+	parser := tvParser2_1{
+		doc: &spdx.Document2_1{},
+		st:  psCreationInfo2_1,
+	}
+	fileName := "f2.txt"
+	err = parser.parsePair2_1("FileName", fileName)
+	err = parser.parsePair2_1("SnippetSPDXID", string(sid1))
+	err = parser.parsePair2_1("PackageName", "p2")
+	if err == nil {
+		t.Errorf("files withoutSpdx Identifiers getting accepted")
+	}
+
+	//case 4 : Invalid File without snippets
+	//Last unpackaged file before the packges start
+	//Last file of a package and New package starts
+	parser3 := tvParser2_1{
+		doc: &spdx.Document2_1{},
+		st:  psCreationInfo2_1,
+	}
+	fileName = "f3.txt"
+	err = parser3.parsePair2_1("FileName", fileName)
+	err = parser3.parsePair2_1("PackageName", "p2")
+	if err == nil {
+		t.Errorf("files withoutSpdx Identifiers getting accepted")
+	}
+	err = parser3.parsePair2_1("PackageName", "p2")
+}
--- a/tvloader/parser2v1/parse_file_test.go
+++ b/tvloader/parser2v1/parse_file_test.go
@@ -886,7 +886,7 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
 }
 
 func TestParser2_1FilesWithoutSpdxIdThrowError(t *testing.T) {
-	// case 1 : The previous file (packaged or unpackaged does not contain spdxID)
+	// case 1: The previous file (packaged or unpackaged) does not contain spdxID
 	parser1 := tvParser2_1{
 		doc:  &spdx.Document2_1{Packages: map[spdx.ElementID]*spdx.Package2_1{}},
 		st:   psFile2_1,
@@ -895,12 +895,11 @@ func TestParser2_1FilesWithoutSpdxIdThrowError(t *testing.T) {
 
 	err := parser1.parsePair2_1("FileName", "f2")
 	if err == nil {
-		t.Errorf("files withoutSpdx Identifiers getting accepted")
+		t.Errorf("file without SPDX Identifier getting accepted")
 	}
 
-	// case 2 : Invalid file with snippet
-	// Last unpackaged file before the snippet start
-	// Last file of a package and New package starts
+	// case 2: Invalid file with snippet
+	// Last unpackaged file before a snippet starts
 	sid1 := spdx.ElementID("s1")
 	fileName := "f2.txt"
 	parser2 := tvParser2_1{
@@ -910,11 +909,11 @@ func TestParser2_1FilesWithoutSpdxIdThrowError(t *testing.T) {
 	}
 	err = parser2.parsePair2_1("SnippetSPDXID", string(sid1))
 	if err == nil {
-		t.Errorf("files withoutSpdx Identifiers getting accepted")
+		t.Errorf("file without SPDX Identifier getting accepted")
 	}
 
 	// case 3 : Invalid File without snippets
-	// Last unpackaged file before the packges start
+	// Last unpackaged file before a package starts
 	// Last file of a package and New package starts
 	parser3 := tvParser2_1{
 		doc: &spdx.Document2_1{},
@@ -927,6 +926,6 @@ func TestParser2_1FilesWithoutSpdxIdThrowError(t *testing.T) {
 	}
 	err = parser3.parsePair2_1("PackageName", "p2")
 	if err == nil {
-		t.Errorf("files withoutSpdx Identifiers getting accepted")
+		t.Errorf("files withoutSpdx Identifier getting accepted")
 	}
 }
--- a/tvloader/parser2v1/parse_file_test.go
+++ b/tvloader/parser2v1/parse_file_test.go
@@ -887,59 +887,57 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
 }
 
 func TestParser2_1FilesWithoutSpdxIdThrowError(t *testing.T) {
-	//case 1
+	// case 1
 	// Last unpackaged file no packages in doc
 	// Last file of last package in the doc
-	var tvPairs []reader.TagValuePair
-	tvPair1 := reader.TagValuePair{Tag: "SPDXVersion", Value: "SPDX-2.1"}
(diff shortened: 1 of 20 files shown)


specter25:
@swinslow , @RishabhBhatnagar sir i have tried to fix the issue #26 , can you please give it a look and suggest if there are any changes :)

tvloader/parser2v1/parse_file_test.go
```diff
@@ -885,4 +886,60 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
 	}
 }
 
+func TestParser2_1FilesWithoutSpdxIdThrowError(t *testing.T) {
+	//case 1
```
RishabhBhatnagar, line 890: [optional] it is often idiomatic to put a space after double slashes for comments.

tvloader/parser2v1/parse_file_test.go
```diff
@@ -885,4 +886,60 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
...
+	// Last file of last package in the doc
+	var tvPairs []reader.TagValuePair
+	tvPair1 := reader.TagValuePair{Tag: "SPDXVersion", Value: "SPDX-2.1"}
+	tvPairs = append(tvPairs, tvPair1)
+	tvPair2 := reader.TagValuePair{Tag: "DataLicense", Value: "CC0-1.0"}
+	tvPairs = append(tvPairs, tvPair2)
+	tvPair3 := reader.TagValuePair{Tag: "SPDXID", Value: "SPDXRef-DOCUMENT"}
+	tvPairs = append(tvPairs, tvPair3)
+	tvPair4 := reader.TagValuePair{Tag: "FileName", Value: "f1"}
+	tvPairs = append(tvPairs, tvPair4)
```
RishabhBhatnagar, line 901: initialization of `tvPairs` can be squashed by using anonymous objects in the definition itself (this change is just for the clarity of expression).

```golang
	tvPairs := []reader.TagValuePair{
		{Tag: "SPDXVersion", Value: "SPDX-2.1"},
		{Tag: "DataLicense", Value: "CC0-1.0"},
		{Tag: "SPDXID", Value: "SPDXRef-DOCUMENT"},
		{Tag: "FileName", Value: "f1"},
	}
```

tvloader/parser2v1/parse_file_test.go
```diff
@@ -885,4 +886,60 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
...
+	_, err := ParseTagValues(tvPairs)
+	if err == nil {
+		t.Errorf("files withoutSpdx Identifiers getting accepted")
+	}
+
+	//case 2 : The previous file (packaged or unpackaged does not contain spdxID)
+	tvPair5 := reader.TagValuePair{Tag: "FileName", Value: "f2"}
+	tvPairs = append(tvPairs, tvPair5)
```
RishabhBhatnagar, line 909: can use anonymous object here as well.

tvloader/parser2v1/parse_package.go
```diff
@@ -20,6 +20,10 @@ func (parser *tvParser2_1) parsePairFromPackage2_1(tag string, value string) err
 	case "PackageName":
 		// if package already has a name, create and go on to a new package
 		if parser.pkg == nil || parser.pkg.PackageName != "" {
+			//check if the previous package contained a spdxId or not
+			if parser.pkg != nil && parser.pkg.PackageSPDXIdentifier == spdx.ElementID("") {
```
RishabhBhatnagar, line 24: [optional] redundant conversion to spdx.ElementID.

RishabhBhatnagar (changes requested):
Note: I haven't yet reviewed the technical aspects of the PR.
I've just mentioned some of the structural changes.

specter25:
@RishabhBhatnagar sir can you pls review it once again .

swinslow:
Hi @specter25, thanks for putting this together and my apologies for being slow to review and respond.

This is good work. I appreciate you preparing tests to cover the different cases. I am going to have a few suggestions and minor requested changes in follow-up comments below. Thanks again for your help with this.

tvloader/parser2v1/parse_file_test.go
```diff
@@ -885,4 +886,58 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
...
+		{Tag: "DataLicense", Value: "CC0-1.0"},
+		{Tag: "SPDXID", Value: "SPDXRef-DOCUMENT"},
+		{Tag: "FileName", Value: "f1"},
+	}
+	_, err := ParseTagValues(tvPairs)
+	if err == nil {
+		t.Errorf("files withoutSpdx Identifiers getting accepted")
+	}
+
+	// case 2 : The previous file (packaged or unpackaged does not contain spdxID)
```
swinslow, line 904: I think case 2 makes sense to keep here. However, I would not structure it using `ParseTagValues()` and a `[]reader.TagValuePair`. Instead, I would do it the way you did for cases 3 and 4 below -- create a `tvParser2_1{}` state object with the start state, and then pass the tag and value to `parser.parsePair2_1()`.

The reason for this is that we want these to be unit tests -- to each be testing as little functionality as possible. If we use ParseTagValues() and `reader.TagValuePair{}`, we have to import the reader package into these tests, and we have broader functionality running in this test than we need.

tvloader/parser2v1/parse_file_test.go
```diff
@@ -885,4 +886,58 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
...
+	tvPairs = append(tvPairs, reader.TagValuePair{Tag: "FileName", Value: "f2"})
+	_, err = ParseTagValues(tvPairs)
+	if err == nil {
+		t.Errorf("%s", err)
```
swinslow, line 908: Note that this won't work -- the test is `if err == nil`, so if we get to the `Errorf()` call then err will be nil and can't be printed. I think you want to replace this with an actual description of what we expected and what we incorrectly got instead?

tvloader/parser2v1/parse_file_test.go
```diff
@@ -885,4 +886,58 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
...
+	// Last unpackaged file before the packges start
+	// Last file of a package and New package starts
+	sid1 := spdx.ElementID("s1")
+	parser := tvParser2_1{
+		doc: &spdx.Document2_1{},
+		st:  psCreationInfo2_1,
+	}
+	fileName := "f2.txt"
+	_ = parser.parsePair2_1("FileName", fileName)
+	_ = parser.parsePair2_1("SnippetSPDXID", string(sid1))
```
swinslow, line 921: For these two lines, instead of running `parsePair2_1` and throwing away the result, can you instead include them in the `tvParser2_1{}` setup above? And then only call parsePair2_1 for the call below that we actually want to test here.

See [here](https://github.com/spdx/tools-golang/blob/2d832aee40b755130804757f02fa1d1f0921b8c8/tvloader/parser2v1/parse_file_test.go#L15-L20) for an example of setting up `tvParser2_1{}` with more fields than just doc and st. You likely wouldn't be using exactly what's there, but I'd suggest following that model.

tvloader/parser2v1/parse_file_test.go
```diff
@@ -885,4 +886,58 @@ func TestParser2_1FailsIfArtifactURIBeforeArtifactName(t *testing.T) {
...
+	err = parser.parsePair2_1("PackageName", "p2")
+	if err == nil {
+		t.Errorf("files withoutSpdx Identifiers getting accepted")
```
swinslow, line 924: Is this what you're actually testing? The line above with `SnippetSPDXID` means that we shifted from parsing a File section to parsing a Snippet section. That point is where the missing File SPDX ID was relevant. If you remove that line, then I think this test would be doing what you want it to do here.

tvloader/parser2v1/parse_package_test.go
```diff
@@ -1069,3 +1070,38 @@ func TestFailsPackageExternalRefWithInvalidFormat(t *testing.T) {
...
+	parser.doc.Packages["p1"] = pkgOld
+	// the Document's Packages should have this one only
+	if parser.doc.Packages["p1"] != pkgOld {
+		t.Errorf("expected package %v, got %v", pkgOld, parser.doc.Packages["p1"])
+	}
+	if len(parser.doc.Packages) != 1 {
+		t.Errorf("expected 1 package, got %d", len(parser.doc.Packages))
+	}
+
+	// Case 2: Checks the Last package
```
swinslow, line 1091: This "Case 2" comment appears to be in the wrong place. The next few lines are still about testing the first case above.

tvloader/parser2v1/parse_package_test.go
```diff
@@ -1069,3 +1070,38 @@ func TestFailsPackageExternalRefWithInvalidFormat(t *testing.T) {
...
+	pkgName := "p2"
+	err := parser.parsePair2_1("PackageName", pkgName)
+	if err == nil {
+		t.Errorf("packages withoutSpdx Identifiers getting accepted")
+	}
+	tvPairs := []reader.TagValuePair{
```
swinslow, line 1097: This appears to be where Case 2 starts. As with the similar code in parse_file_test.go, I think this belongs in parser_test.go instead. It looks like this is actually testing code that occurs at the end of parsing via parser.go. Can you move this out into parser_test.go?

tvloader/parser2v1/parse_snippet.go
```diff
@@ -34,6 +34,10 @@ func (parser *tvParser2_1) parsePairFromSnippet2_1(tag string, value string) err
 	// tag for creating new package section and going back to parsing Package
 	case "PackageName":
 		parser.st = psPackage2_1
+		// check here whether the last file of the previous package contained the FileSpdxIdentifier
```
swinslow, line 37: When parsing Snippets, I don't think you want to do the check here in the `PackageName` tag. Rather, `SnippetSPDXID` is what signals the end of the previous File section and the beginning of the current Snippet section.

Because of that, the section of `parse_snippet.go` that handles the `SnippetSPDXID` tag is where I think this check should go.

Finally, also note the comments earlier about communicating to the user via the error message about which File's filename we're talking about.

swinslow: Also, sorry, I meant to ask -- is there a test corresponding to this code? I don't see any additions made to parse_snippet_test.go.

tvloader/parser2v1/parse_package.go
```diff
@@ -20,6 +20,10 @@ func (parser *tvParser2_1) parsePairFromPackage2_1(tag string, value string) err
 	case "PackageName":
 		// if package already has a name, create and go on to a new package
 		if parser.pkg == nil || parser.pkg.PackageName != "" {
+			// check if the previous package contained a spdxId or not
+			if parser.pkg != nil && parser.pkg.PackageSPDXIdentifier == nullSpdxElementId2_1 {
+				return fmt.Errorf("invalid package without a package SPDX identifier")
```
swinslow, line 25: Similar to the File error messages, can you tell the user in this error message what is the Package's PackageName, so that they can tell which one is incorrect?

tvloader/parser2v1/parser.go
```diff
@@ -20,6 +20,12 @@ func ParseTagValues(tvs []reader.TagValuePair) (*spdx.Document2_1, error) {
 			return nil, err
 		}
 	}
+	if parser.file != nil && parser.file.FileSPDXIdentifier == nullSpdxElementId2_1 {
```
swinslow, line 23: Same comments here and for the next few lines, tell the user which file name / Package name is causing the problem.

tvloader/parser2v2/parse_creation_info.go
```diff
@@ -12,7 +12,7 @@ import (
 func (parser *tvParser2_2) parsePairFromCreationInfo2_2(tag string, value string) error {
 	// fail if not in Creation Info parser state
 	if parser.st != psCreationInfo2_2 {
-		return fmt.Errorf("Got invalid state %v in parsePairFromCreationInfo2_2", parser.st)
+		return fmt.Errorf("got invalid state %v in parsePairFromCreationInfo2_2", parser.st)
```
swinslow, line 15: I'm not going to reproduce all of them, but I think all of my comments for the 2_1 code above also apply to the 2_2 code here, and should be handled similarly.

swinslow (changes requested):
Thanks again @specter25 for your help. I know there are a lot of comments here, but I think they will help fix some issues and make for cleaner and more maintainable code.

tvloader/parser2v1/parse_file_test.go
specter25: Sir these two lines are needed as the logic requires the parser.file to point towards the last parsed file (which is invalid)

specter25: @swinslow

specter25:
@swinslow sir , I have done the changes from my end . You can review the changes once :) sorry for the late response .

swinslow:
Hi @specter25, good work on this one. I just pushed a few minor changes to the text for comments / messages, but otherwise looks good -- merging now. Thanks!

swinslow (approved):

Issue #71, opened 2021-06-10
Add utility function for retrieving packages with arbitrary relationships


ianling:
https://github.com/spdx/tools-golang/blob/main/spdxlib/described_elements.go

This file contains functions that allow you to retrieve the IDs of the packages described by the document.

Would it be possible to get a similar function that allows us to retrieve IDs of packages that have a given relationship, e.g. DEPENDS_ON/DEPENDENCY_OF, rather than specifically DESCRIBES/DESCRIBED_BY?

Something like:

```
func GetPackagesByRelationship(doc *spdx.Document2_2, relationship string, ref spdx.ElementID) ([]spdx.ElementID, error) { ... }

usage:
GetPackagesByRelationship(doc, "DEPENDENCY_OF", packageA)
GetPackagesByRelationship(doc, "DEPENDS_ON", packageB)
GetPackagesByRelationship(doc, "DESCRIBED_BY", document)
```

swinslow:
This is a great suggestion, thanks @ianling. Yes, I expect this would be very useful.

One consideration would be whether the search should be on just one side of the relationship (e.g. just RefA or just RefB), or should search both. Most of the [relationship types](https://spdx.github.io/spdx-spec/7-relationships-between-SPDX-elements/#71-relationship) are "directional" in that it's meaningful which is on the left vs. right side, but some I think probably are not (e.g. DYNAMIC_LINK, STATIC_LINK).

Perhaps part of the function call should be to specify whether to search just left side, just right side, or both?

Either way, yes, I'm a big fan of adding this. If you'd like to propose a PR, please feel free!

swinslow:
Fixed in #74, thank you @ianling!

PR #73, opened 2021-06-15
Add relationship filter function

ianling:
Addresses #71 

Adds a function for filtering relationships to retrieve an arbitrary set of ElementIDs. This can be seen in action in described_elements.go:

```
	eIDs, err := FilterRelationships2_1(doc, func(relationship *spdx.Relationship2_1) *spdx.ElementID {
		refDocument := spdx.MakeDocElementID("", "DOCUMENT")

		if relationship.Relationship == "DESCRIBES" && relationship.RefA == refDocument {
			return &relationship.RefB.ElementRefID
		} else if relationship.Relationship == "DESCRIBED_BY" && relationship.RefB == refDocument {
			return &relationship.RefA.ElementRefID
		}

		return nil
	})
```

This retrieves a slice of all the ElementIDs that are described by the document, accounting for both DESCRIBES and DESCRIBED_BY relationship types.

This function was intended to be generic so that additional utility functions like `GetDescribedPackageIDs2_1` can be added more easily, and with less code duplication.

One slight paradigm shift here is that I removed the document validity checks from the `GetDescribedPackageIDs2_*` functions, in favor of having separate `ValidateDocument` functions, and assuming that the user has run those first before attempting to dig into the document. If you think this is a bad idea, let me know and I will remove it. My goal was to prevent every utility function from having to include a bunch of repetitive document validation.


ianling:
Hang on... figuring out what the deal is with this PR

ianling:
Closed in favor of #74

PR #94, opened 2021-07-24
[GSoC'21] Pr:6 . JSON saver tests

specter25:
GSoC'2021 Pull Request 5 :
Deliverables Completed :

- [x] complete unit tests on Jsonsaver
- [x] fix issues raised in Gsoc pr 5 related to Jsonsaver


specter25:
cc @swinslow I have made all the changes that you suggested in the previous meeting and moreover complete the unit tests as well as the complete functionality tests

jsonsaver/jsonsaver_test.go
```diff
@@ -0,0 +1,230 @@
...
+				LicenseConcluded:  "(LGPL-2.0-only OR LicenseRef-2)",
+				LicenseInfoInFile: []string{"GPL-2.0-only", "LicenseRef-2"},
+				FileNotice:        "Copyright (c) 2001 Aaron Lehmann aaroni@vitelus.",
+			},
+		},
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantW   string
```
swinslow, line 200: Is `wantW` being used here? I see that `wantErr` is being assigned and checked below, but it looks like this isn't actually checking to see whether the output of `Save2_2` matches what's expected.

That might be intentional, since we can't be certain about the exact output of bytes from the JSON marshaller. And I believe you're checking the actual contents of the saver output in the other tests in the subdirectory. If that's the case, though, then perhaps it makes sense to remove `wantW` here to avoid confusion?

jsonsaver/jsonsaver_test.go
```diff
@@ -0,0 +1,230 @@
...
+		wantErr bool
+	}{
+		// TODO: Add test cases.
```
swinslow, line 203: Here and for the other test files below -- are you still planning to add more test cases? If so, fine to leave this in, but if not then once you're done I'd suggest removing these comments.

jsonsaver/saver2v2/save_creation_info_test.go
```diff
@@ -0,0 +1,90 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := renderCreationInfo2_2(tt.args.ci, tt.args.jsondocument); (err != nil) != tt.wantErr {
+				t.Errorf("renderCreationInfo2_2() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			for k, v := range tt.want {
+				if !reflect.DeepEqual(tt.args.jsondocument[k], v) {
+					t.Errorf("Load2_2() = %v, want %v", tt.args.jsondocument[k], v)
```
swinslow, line 85: Should `Load2_2()` be `renderCreationInfo2_2()` in this comment instead?

swinslow (changes requested):
Hi @specter25, great work on this PR as always. I've got a few minor comments on particular bits which should show up in this review.

Unfortunately, there's one larger issue with the tests that I'm seeing which is pretty significant, and is randomly causing the tests to fail from time to time when I re-run them (even though they didn't fail in the CI when this PR was submitted). I'll explain below what the issue is. Because it's causing tests to fail, I'd suggest addressing it before I merge this PR into the `json` branch. Take a look at my comments below and let me know if you have questions!

= = =

The problem I'm seeing is that, when I re-run the tests repeatedly, some of them are failing from time to time (though sometimes they are passing). You can re-run the tests on your local machine by running `go clean -testcache` to clear the testing cache, followed by `go test ./...` as usual.

The problems I'm seeing appear to be where the items in a slice are getting saved into different orders on each test run. The test code expects them to appear in a specific order that you define in the test; and if they are in that order, then it passes, but sometimes they are in a different order.

Specifically, it's where the items are stored in a map in the tools-golang SPDX data model, but in an array (or slice) in the SPDX JSON format. Because of the difference in data structures, you are (correctly) doing a `for ... range` over the map, and then `append`ing the results into a slice to get rendered in JSON as an array.

The problem is that when doing `for ... range` over a map, Golang returns the map contents in a random order. You can see [this article](https://nathanleclaire.com/blog/2014/04/27/a-surprising-feature-of-golang-that-colored-me-impressed/) for more details, or [this shorter one (under 'Iteration order')](https://blog.golang.org/maps) as well.

So because the `for ... range` call returns the map contents in a random order, they will be randomly ordered into the slice as well. And that means that when you eventually do the `DeepEqual` call to check the results, they will sometimes pass and sometimes fail, depending on the random result.

Here's a couple of specific areas where I'm seeing this error:
* when rendering all Files into the "files" array: https://github.com/spdx/tools-golang/blob/81c7677913cec65e6f45f8def531783b3399b0bc/jsonsaver/saver2v2/save_files.go#L12
* when rendering a file's checksums: https://github.com/spdx/tools-golang/blob/81c7677913cec65e6f45f8def531783b3399b0bc/jsonsaver/saver2v2/save_files.go#L29

To be clear, there could be others; those are just the ones I'm seeing so far. Basically, anywhere in the saver code where you are doing a `for ... range` over a map and storing the results into a slice, I'm guessing this could occur.

I think there's a couple of ways you could go about fixing this.

One option would be in the saver code: every time you create a slice from a map, always make sure the slice is sorted in the same order. This could be done for example by doing the `for ... range` over the map just to get the keys; then sorting the keys; and then iterating over the keys to get the values. [The blog post I mentioned above](https://blog.golang.org/maps) explains how to do this, under the "Iteration order" section. If you do this, then the output from the render functions should always be consistent, and then the `DeepEqual` calls can always check for that consistent output.

A second option could be to have the `DeepEqual` call somehow ignore the ordering of slices when it compares the result to the expected structure. I'm not sure whether there's a straightforward way to do that; and it might not be a good idea anyway, in case there are any instances where ordering in slices actually does matter. (I don't know of any for SPDX at the moment.)

So I'd lean towards the first option, though if you have another thought, I'm happy to hear it!

swinslow:
Hi @specter25, from our separate chat I understand that you're planning to submit a subsequent PR to handle the testing issue I mentioned above. So I'm going ahead and merging this one into the json branch. Thanks!

PR #99, opened 2021-08-16
Refactor error returns jsonloader

josedonizetti:
Minor change to refactor errors returned, the idea is to only use `fmt.Errorf` if there is a format. If not it either returns the error directly, or use `errors.New` to create it.

Let me know if you okay with simple PRs like this. Thanks


swinslow:
Thank you @josedonizetti !

@specter25, you're closest to the jsonloader code -- can you take a look at this and add a review?

PR #100, opened 2021-11-25
fixing bug in jsonsaver

yanyag:
Avoid producing invalid json with "Files: null" and "Snippets: null" if saving a package without files (filesAnalyzed is false)

Signed-off-by: Yan Yagudayev <yanyag@gmail.com>

### DIFF
--- a/jsonsaver/saver2v2/save_document.go
+++ b/jsonsaver/saver2v2/save_document.go
@@ -71,7 +71,7 @@ func RenderDocument2_2(doc *spdx.Document2_2, buf *bytes.Buffer) error {
 	}
 
 	// save files and snippets from spdx to json
-	if allfiles != nil {
+	if len(allfiles) > 0 {
 		_, err = renderFiles2_2(doc, jsondocument, allfiles)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


yanyag:
@swinslow, 
Could you please review and approve this small bug fix?

swinslow:
Hi @yanyag, thanks very much for this -- and my apologies for the slow response, as I haven't turned back to the SPDX Golang tools for a while.

I've created an issue for this at #103. I think the PR you shared here is a good start, though I'm not sure it sufficiently solves the issue. For instance, if an SPDX Document contains some Files but no Snippets, then `snippets: null` will still be outputted. It also looks like there are some other properties (e.g. `externalDocumentRefs`) where the same null issue is happening.

I'll dig a bit deeper into the jsonsaver code to see how best to address this.

Separately, I noted the change you proposed in this PR to save_document_test.go. Can you help clarify for me what the changed / additional tests are meant to add? It wasn't immediately obvious to me, but I might need to stare at them a bit more...

swinslow:
Hi @yanyag, the remaining issues I mentioned in the prior comment were resolved with #106, so I'm closing this PR. Thank you for raising the issue!

CatalinStratu:
Hi, can I handle this issue?
I really want to participate in GSOC in this project and the more issues I solve, the better I will understand the project!

yanyag:
Thanks a lot @swinslow

Issue #69, opened 2021-06-05
Rename primary branch to 'main'


swinslow:
see https://github.com/github/renaming#renaming-existing-branches

swinslow:
done, see https://github.com/spdx/tools-golang/branches

swinslow:
For anyone who needs to change their local branch settings, here are the instructions from GitHub (you should see these as well when you visit the repo page):

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

Note that you may need to change `origin` to e.g. `upstream` depending on your local settings.

PR #112, opened 2022-03-13
Fix pulling main branch for code coverage

swinslow:
After the branch renaming from 'master' to 'main' in 0.2.0 per #69,
the tests were being run by GitHub Actions on the 'main' branch.
However, the code coverage tooling appears to still have been
pulling the old 'master' branch and therefore was not including
later changes such as the new JSON code. This PR fixes this by
changing the branch name in the GitHub Actions config file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ jobs:
   tests:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


swinslow:
Note for future reference: It looks like the change that was needed was going into https://coveralls.io/github/swinslow/tools-golang/settings and re-syncing with GitHub to pull the updated default branch info. Still, this PR's change was worth making anyway  :)

PR #114, opened 2022-03-14
builder: File paths are not relative, bug fixes

CatalinStratu:
After my changes: 
image

Old:
image

Also, I've added a comment containing the command that generates the documentation to make it easier to test.
Have a nice day!!


swinslow:
Hi @CatalinStratu, thanks for this! It looks like it's in the right direction, I have a couple of comments below:

1. Just to clarify, the `FileName` file paths should be relative to the _root of the package being scanned_, rather than relative to the _place where the program is running_. For example, if we have the following at `tools-golang/testdata/project1/`:

```
testdata
├── project1
│   ├── emptyfile.testdata.txt
│   ├── file1.testdata.txt
│   ├── file3.testdata.txt
│   ├── folder1
│   │   └── file4.testdata.txt
│   ├── lastfile.testdata.txt
. . .
```

and we're running a program which is at `tools-golang/examples/3-build/`, then when we're describing the Files in the `project1` Package, the FileNames should be:
```
./emptyfile.testdata.txt
./file1.testdata.txt
./file3.testdata.txt
./folder1/file4.testdata.txt
. . .
```

rather than `./../../testdata/project1/emptyfile.testdata.txt`, etc. Does that make sense?

2. I agree with @seabass-labrax above, it's a great addition to have the comment in the `examples/` source code for each example program explaining how to run it. I'm okay with you leaving in this PR the one you added for example 3, but if so then could you also submit a separate issue so that we remember to add something similar for the others? Thanks!

CatalinStratu:
@swinslow  Hi, "spdx / tools-golang" behaves differently on Windows than on Linux. I made some changes and for Linux now it works perfectly, I'm making some changes so that it works well on Windows as well.

swinslow:
Thank you @CatalinStratu! I'll review the updates later today and will merge if it all looks good to go, or will circle back with any follow-up comments or questions.

CatalinStratu:
@swinslow I added 'if osType == "windows"' because in Linux the folder looks good, in Windows it adds "/../../". After this PR I created a new PR explaining how to run the examples

CatalinStratu:
Hi @swinslow,  what do you think of this PR??

swinslow:
Hi @CatalinStratu, apologies for the delayed response. I am reviewing now and will merge or else circle back shortly with any further thoughts.

swinslow:
Although I haven't yet been able to test this myself on Windows, I'm not seeing any issues based on my tests on Linux. So I'm going to go ahead and merge this. @CatalinStratu thanks for your help and patience!

PR #129, opened 2022-04-01
Cross Platform Testing TestBuilder2_1CanBuildPackageSection

CatalinStratu:
I propose to do individual tests for Linux and Windows. `"/testdata/project1_windows"` has no symbolic link because I create it from code. To run the tests on **Windows is needed to run the tests as an administrator**

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ Go programs to work with [SPDX®](https://spdx.dev/) files.
 
 ## Recent news
 
-2022-03-27: **v0.3.0-rc1**: added support for saving SPDX JSON files as well as
+2022-04-03: **v0.3.0**: added support for saving SPDX JSON files as well as
 other improvements and bugfixes. See [RELEASE-NOTES.md](./RELEASE-NOTES.md)
-for full details (to be updated for v0.3.0 prior to final release).
+for full details.
 
 ## What it does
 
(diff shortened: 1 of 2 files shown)


swinslow:
Thanks @CatalinStratu. Before merging this, I'm going to want to do a bit more research into how Git on Windows vs. Linux handles symbolic links.

There's also an open question in [the main SPDX specification repo](https://github.com/spdx/spdx-spec/issues/610) about how symlinks should be handled when creating SPDX documents. Until that issue is resolved, I don't think we'll have a definite answer on how best to handle this.

Instead, I would suggest for the time being let's skip those two particular test cases when the test suite is run on Windows. I'd prefer we skip them, rather than requiring the test suite to run with admin rights (since users rightfully might hesitate to do so).

We'll want to open an issue to track this symlinks question specifically, and then leave it open until the main question in the spdx-spec issue is resolved. In the meantime we can just skip those two failing tests when running on Windows so that we can move forward with setting up the CI for all the other tests. Let me know what you think. Thanks!

builder/builder2v1/build_package.go
```diff
@@ -40,7 +40,7 @@ func BuildPackageSection2_1(packageName string, dirRoot string, pathsIgnore []st
 	for _, fp := range filepaths {
 		newFilePatch := ""
 		if osType == "windows" {
-			newFilePatch = filepath.FromSlash("." + fp[dirRootLen:])
+			newFilePatch = filepath.FromSlash(fp[dirRootLen:])
```
CatalinStratu, line 43: "./"

Issue #124, opened 2022-03-28
tvloader: Error when Snippet does not immediately follow its File section


CatalinStratu:
When two identical files are inserted, the error below appears. 
![image](https://user-images.githubusercontent.com/28156781/160440763-92b4da6e-4923-4bb6-96bf-09d3a6de5250.png)

The file that I used when I identified this error is: [https://github.com/spdx/spdx-spec/blob/development/v2.2.2/examples/SPDXTagExample-v2.2.spdx](https://github.com/spdx/spdx-spec/blob/development/v2.2.2/examples/SPDXTagExample-v2.2.spdx)

I want to mention that the SnippetSPDXID tag is in both files.

swinslow:
Thanks for filing this @CatalinStratu -- I'm taking a look now to see whether this is an issue with the example / sample document, or with the tvloader parser.

As mentioned [here](https://github.com/spdx/tools-golang/pull/123#discussion_r839067848), note though that I think the files being compared in this example are not actually identical. That shouldn't cause the parsing issue here, but just flagging to avoid confusion.

swinslow:
OK -- so it looks like this is a parsing error in tvloader, due to an ambiguity in the SPDX spec. See https://github.com/spdx/spdx-spec/issues/651 for more details.

Basically, one possible reading of the spec says that, for tag-value files, a Snippet section should always come immediately after the File section that "contains" it. But a different reading is just that the Snippet needs to be _somewhere_ after the corresponding File, even if it's not immediately after it.

The current tvloader code takes the first reading, and because of that it assumes that a `SnippetSPDXID` tag will only appear when it's in the middle of parsing a File section. So in the sample file at `examples/sample-docs/tv/SPDXTagExample-v2.2.spdx`, it hits a parsing failure when it encounters the `SnippetSPDXID` tag coming after a Package section.

We'll need to see how https://github.com/spdx/spdx-spec/issues/651 is resolved before we can address this here. That means we'll kick this down the road to 0.4.0.

In the meantime, for the 0.3.0 release I'm going to add another different sample document, so that example 6 can be run without erroring out.

CatalinStratu:
@swinslow I would like to work on this issue, can you give me some suggestions on where I can look for the problem? Thanks!

CatalinStratu:
Hi @swinslow, this problem was solved by @ianling in the PR #133

CatalinStratu:
This problem was solved in the PR #133

PR #132, opened 2022-04-11
WIP: Refactor JSON parser/writer, add YAML support

ianling:
Refactors the JSON parser/writer so it no longer uses `reflect`. The new approach is incomparably more maintainable and less error-prone, and requires substantially fewer unit tests.

Adds YAML support.

Using Go's built-in JSON functionality gives us YAML support for free using https://pkg.go.dev/sigs.k8s.io/yaml

There are a number of breaking API changes here (I changed some of the structs to align more with the spec). Maybe once full support for all three of these file types is finished in this PR (plus RDF, which is already in the module), this warrants an API-breaking 1.0.0 release of these tools?

I'm investigating XLS support as well, which, if added, would mean that this module has full support for all of the formats listed in the spec.


ianling:
(title of PR originally mentioned XML because I got this mixed up with CycloneDX, where XML is already a supported format)

CatalinStratu:
Hi, @ianling, it looks great, we enjoy new contributors.

I believe that this PR should be divided into smaller PRs to make it easier to code review.

CatalinStratu:
Hi, @ianling, it looks great, we enjoy new contributors.

I believe that this PR should be divided into smaller PRs to make it easier to code review.

ianling:
I am not sure how to break it into smaller chunks unfortunately; in order to maintain the same functionality in the JSON parser/writer, the entire struct refactor needs to be added at once. If I do it piecemeal, I have to go back and make changes to the old JSON parser/writer which would then be removed in my later commits.

CatalinStratu:
In this PR, issue #124 is fixed, but there is a new issue
image

CatalinStratu:
> I am not sure how to break it into smaller chunks unfortunately; in order to maintain the same functionality in the JSON parser/writer, the entire struct refactor needs to be added at once. If I do it piecemeal, I have to go back and make changes to the old JSON parser/writer which would then be removed in my later commits.
> 
> In other words, if I break it into smaller commits, there will be commits where the tools are broken.
> 
> I can move the YAML commit into a different PR, but the changes for that are contained entirely in the `yaml` folder.

@ianling ,yes, it is a good idea to move YAML to PR separately

ianling:
I missed running those examples during my testing, I will do that and report back (and fix bugs I introduced). Thanks for taking a look.

ianling:
Closing this PR in favor of !133 and !134

PR #134, opened 2022-04-11
YAML support

### LINKED ISSUE #135: YAML Parser and Writer
Just adding issues to track work I am doing downstream.

YAML support is added in #134, but this PR relies on #133, as it builds off the top of the JSON refactor

ianling:
Depends on (and contains the commit from) https://github.com/spdx/tools-golang/pull/133. Closes #135 

Adds YAML support.

Note to self: rebase before undrafting

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ tools-golang provides the following packages:
 * *tvsaver* - tag-value document saver
 * *rdfloader* - RDF document loader
 * *json* - JSON document parser and writer
+* *yaml* - YAML document parser and writer
 * *builder* - builds "empty" SPDX document (with hashes) for directory contents
 * *idsearcher* - searches for [SPDX short-form IDs](https://spdx.org/ids/) and builds SPDX document
 * *licensediff* - compares concluded licenses between files in two packages
--- a/go.mod
+++ b/go.mod
@@ -5,4 +5,5 @@ go 1.13
 require (
 	github.com/google/go-cmp v0.5.7
 	github.com/spdx/gordf v0.0.0-20201111095634-7098f93598fb
+	sigs.k8s.io/yaml v1.3.0
 )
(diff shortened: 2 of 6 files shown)


ianling:
@swinslow this one is next (they go in the order they were created in, to clarify). I'll rebase it really quick so it loses the commit from the previous PR

swinslow:
Thanks @ianling! So, I've got one annoying license-related question about this:

Looking through the new sub-dependencies, it looks like they're a mix of MIT, BSD-2-Clause, BSD-3-Clause and ISC, all of which are permissive and typically seen as compatible with either Apache-2.0 OR GPL-2.0-or-later (and therefore I'm fine with them).

But I notice that [gopkg.in/yaml.v2](https://github.com/go-yaml/yaml/tree/v2.4.0) appears to be primarily [Apache-2.0](https://github.com/go-yaml/yaml/blob/v2.4.0/LICENSE). While that's obviously compatible with the Apache-2.0 choice for the SPDX Golang tools, Apache-2.0 is viewed by some as being [incompatible with GPL-2.0](https://www.gnu.org/licenses/license-list.en.html#apache2) if someone uses that choice instead.

My question is, is gopkg.in/yaml.v2 likely going to be a required part of any YAML library in the Golang ecosystem (such that there isn't really an easy alternative)? If so, then we can just make it clear that the YAML functionality for the Golang tools is Apache-2.0 OR GPL-3.0-or-later rather than GPL-2.0.

Sorry for the complication. Let me know your thoughts and if you think we'll need to stick with gopkg.in/yaml.v2, then we can do that and adjust the license details accordingly. Thanks!

ianling:
Unfortunately, it looks like there is pretty much only one other option for a YAML library (https://github.com/goccy/go-yaml/blob/master/go.mod), and one of its transitive dependencies is actually the YAML library used here... so I think that no matter what, in order to enable YAML support, we will need to import that library.

I appreciate you looking into all the license complexities, I'm not very familiar with the compatibilities of the different licenses.

swinslow:
Gotcha, sounds good! And no worries. I'll add a separate issue to track the new dependencies and licenses, and I'll update the related text in the README file once the rest of this all lands. Merging this one now. Thanks again!

PR #137, opened 2022-04-18
WIP: Implement Stringer interface on structs, add FromString method where relevant

ianling:
Allows structs to be reliably and consistently converted to strings (e.g. for tag-value and spreadsheet export/writing)

Allows structs to be filled in by parsing them from strings (e.g. for tag-value and spreadsheet import/parsing)

Depends on and contains commits from #133 and #134


ianling:
@swinslow this one's up next. The methods added here should also probably be used in the tag-value and RDF parsers/writers, for consistency.

spdx/identifier.go
```diff
@@ -14,13 +14,63 @@ import (
...
+
+// FromString parses an SPDX Identifier string into an ElementID.
+// These strings take the form: "SPDXRef-some-identifier"
+func (e *ElementID) FromString(idStr string) error {
+	idFields := strings.SplitN(idStr, "SPDXRef-", 2)
+	switch len(idFields) {
+	case 2:
+		// "SPDXRef-" prefix was present
+		*e = ElementID(idFields[1])
+	case 1:
```
swinslow, line 39: I'm not sure I follow this case: if the `SPDXRef-` prefix is not present, I think we'd want to treat it as an invalid ID rather than ignore that the prefix isn't present? Unless there's a use case here I'm misunderstanding.

spdx/identifier.go
```diff
@@ -66,37 +125,57 @@ func (d *DocElementID) UnmarshalJSON(data []byte) error {
...
+
+	return json.Marshal(d.String())
+}
+
+// MakeDocElementID takes strings for the DocumentRef- and SPDXRef- identifiers (these prefixes will be stripped if present),
+// and returns a DocElementID.
+// An empty string should be used for the DocumentRef- portion if it is referring to the present document.
 func MakeDocElementID(docRef string, eltRef string) DocElementID {
+	docRef = strings.Replace(docRef, "DocumentRef-", "", 1)
+	eltRef = strings.Replace(eltRef, "SPDXRef-", "", 1)
```
swinslow, line 177: I'm a bit uncomfortable with having this function be able to take strings which _might or might not_ have the `DocumentRef-` or `SPDXRef-` prefixes.

This is an edge case probably, but in an SPDX Document you could have two different SPDX elements with the following SPDX Identifiers:
* `SPDXRef-File1`
* `SPDXRef-SPDXRef-File1`

I wouldn't recommend anyone actually build an SPDX Document this way, but I think it would be syntactically valid.

So if we've got functions that are optionally stripping the first instance of `SPDXRef-` that it sees, I feel like that will lead to unexpected behavior for edge cases like the one above. I'm having trouble putting my finger on a concrete example, but I guess I'm thinking that it'll be easier to avoid doing the wrong thing if a function like this always requires its inputs to either have the prefix or omit the prefix.

spdx/package.go
```diff
@@ -15,22 +15,29 @@ type Supplier struct {
...
+// FromString parses a string into a Supplier.
+// These stings take the form: "<SupplierType>: <Supplier>"
+func (s *Supplier) FromString(value string) error {
+	if value == "NOASSERTION" {
+		s.Supplier = value
 		return nil
 	}
 
-	supplierFields := strings.SplitN(supplierStr, ": ", 2)
+	supplierFields := strings.SplitN(value, ": ", 2)
```
swinslow, line 37: Here and for other similar instances elsewhere in the PR, same question as above about whether it's useful to leverage the `extractSubs()` function from the tag-value loader. If not, that's fine, but I'm guessing it may need to handle the possibility of more than 1 whitespace character after the colon.

tvloader/parser2v1/parse_creation_info.go
```diff
@@ -102,7 +102,7 @@ func (parser *tvParser2_1) parsePairFromCreationInfo2_1(tag string, value string
 
 // ===== Helper functions =====
 
-func extractExternalDocumentReference(value string) (string, string, string, string, error) {
+func extractExternalDocumentReference(value string) (spdx.DocElementID, string, string, string, error) {
```
swinslow, line 105: This might be worth keeping as string, given the comments earlier above.

swinslow (changes requested):
Hi @ianling, thanks for this and apologies again for my delay in reviewing it.

I have a few comments below, mostly around how this handles strings that might have extra whitespace and whether to leverage some of the tag-value parsing functions that I think already address some of these. Let me know your thoughts. Thanks!

spdx/identifier.go
ianling: I think it might be useful for this library to accept slightly malformed data and sorta "clean it up" if the user chooses to export it again. In this case, if the user parsed a file and an SPDXID field was missing the `SPDXRef-` prefix, they could export the file again after parsing it and the file would then pass validation (assuming that was the only issue with the file).

If that happens though, we should probably at least issue a warning rather than just charging ahead, or force the user to opt out of validation in order to parse a document that contains problems like this.

I'll give this one some more thought.

spdx/identifier.go
ianling: I think your example should just work, due to the `, 1)` at the end of the `strings.Replace(...)` call.

```
strings.Replace("SPDXRef-File1", "SPDXRef-", "", 1) -> "File1"
strings.Replace("SPDXRef-SPDXRef-File1", "SPDXRef-", "", 1) -> "SPDXRef-File1"
strings.Replace("SPDXRef-DocumentRef-File1", "SPDXRef-", "", 1) -> "DocumentRef-File1"
strings.Replace("File1-SPDXRef", "SPDXRef-", "", 1) -> "File1-SPDXRef"
```

If we can come up with edge cases we're worried about, I can add test cases for them to ensure that this function is handling them the way we'd want.

But as I mentioned in another comment, this is a situation we have run into repeatedly with real SPDX documents from users, where sometimes they just omit the `SPDXRef-` prefix on their identifiers, and it's not sustainable for us to fix that manually every time it happens. I think this is a case where the user clearly meant to add the prefix but just forgot/missed it and it's safe for us to make an assumption.

spdx/checksum.go
```diff
@@ -24,3 +29,34 @@ type Checksum struct {
 	Algorithm ChecksumAlgorithm `json:"algorithm"`
 	Value     string            `json:"checksumValue"`
 }
+
+// FromString parses a Checksum string into a spdx.Checksum.
+// These strings take the following form:
+// SHA1: d6a770ba38583ed4bb4525bd96e50461655d2759
+func (c *Checksum) FromString(value string) error {
```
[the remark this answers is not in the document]
ianling: Woops, I didn't realize those functions existed. I have refactored this PR to move those functions up so they're not buried in multiple utils folders, so now we consistent behavior across all the different file parsers.

spdx/document.go
```diff
@@ -33,6 +38,92 @@ type ExternalDocumentRef2_2 struct {
...
+	return fmt.Sprintf("%s %s %s", e.DocumentRefID, e.URI, e.Checksum)
+}
+
+// String converts a ExternalDocumentRef2_2 object to a string.
+// These strings take the form: "<DocumentRefID> <URI> <Checksum>"
+func (e ExternalDocumentRef2_2) String() string {
+	return fmt.Sprintf("%s %s %s", e.DocumentRefID, e.URI, e.Checksum)
+}
+
+// FromString parses a string into a spdx.ExternalDocumentRef2_1.
```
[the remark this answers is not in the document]
ianling: Moved this `extract...` function up out of that folder as well so we're using it everywhere now.

spdx/package.go
ianling: This PR now uses the code from `extractSubs` instead of my implementation that only supported one space.

tvloader/parser2v1/parse_creation_info.go
ianling: Resolving this comment to keep discussion about string vs DocElementID constrained to one comment

ianling:
Thank you again for your time @swinslow. I moved some code up from the `utils` folders so that we are using the same logic in all the parsers.

rdfloader/parser2v2/utils_test.go
```diff
@@ -184,106 +151,3 @@ func Test_getNodeTypeFromTriples(t *testing.T) {
 		t.Errorf("expected an error saying more than one rdf:type found, got %v", err)
 	}
 }
-
-// following tests are copy pasted from tvloader/parser2v2/util_test.go
-
-func TestCanExtractDocumentAndElementRefsFromID(t *testing.T) {
```
ianling, line 190: I made sure to hang on to non-duplicate tests and move them to the proper place after my re-arranging, so we did not lose any coverage.

kzantow:
Hi yet again @ianling -- are you interested in updating this PR based on the latest refactoring?

PR #139, opened 2022-04-21
Support more checksum algorithms

ianling:
Depends on and contains commits from #134, #135, #137, and #138 

Allows all the known checksum algorithms in the enum to be considered valid.

Fixes #96

PR #138, opened 2022-04-18
Spreadsheet / XLSX parser and writer

ianling:
Adds a spreadsheet/XLSX/XLS parser and writer. Closes #136 

Includes unit tests:

1. Parses the example XLSX file from here: https://github.com/spdx/spdx-spec/tree/development/v2.2.2/examples and then checks the result against a handwritten Document struct
2. Writes out the handwritten struct to a XLSX file, and then parses the written XLSX file to verify that no data was lost and all the values match what was expected.

Depends on and contains commits from #133, #134, and #137

Also bumps the Github Actions workflow to use Go 1.15, because that is the minimum version supported for the Excelize module used for interacting with XLSX files.

PR #147, opened 2022-06-06
Make FilesAnalyzed a pointer, assume "true" if absent

### LINKED ISSUE #145: Bug: Package FilesAnalyzed field does not behave according to spec
Per the spec, if the FilesAnalyzed field is absent from a document, that should be interpreted as `FilesAnalyzed: true`.

Unfortunately, for the JSON and YAML parsers/savers, this was missed, so if they are absent, they default to false.

Additionally, the zero/empty value of a boolean in Go is `false`, and the FilesAnalyzed field's JSON tag contains `omitempty`, so we are unable to write out the FilesAnalyzed field to a file if it is false; it will simply be omitted, which results in other things (like the online SPDX validator tool) interpreting it as `true`.

We should change this field to a pointer to denote that it can be absent, and also make sure that we handle that case properly by interpreting it as `true` across all the different file parsers.

ianling:
Closes #145 

Depends on and contains commits from https://github.com/spdx/tools-golang/pull/134, https://github.com/spdx/tools-golang/issues/135, https://github.com/spdx/tools-golang/pull/137, https://github.com/spdx/tools-golang/pull/138, and #139 

Per the spec, FilesAnalyzed is not required to be present. If it is not present, the value is assumed to be `true`.

This was previously handled using a secondary field that denoted whether or not FilesAnalyzed was present.

This PR removes that field and makes FilesAnalyzed a pointer to denote that it is not required to be present.

All of the parsers interpret an absence of a value as `true`.

All of the file writers will ALWAYS write a value for FilesAnalyzed, either true or false, they will never omit it. I think it is better for SPDX files to be explicit.


kzantow:
Hi @ianling -- there was a significant refactoring in this library, are you able to adjust this PR for the changes?

kzantow:
Hiya @ianling -- thank you for this (and the other) contributions. Since this is dependent on the others, I'm going to close it for now. We will gladly reopen this if you decide to update/continue it, but I think it also may have been superseded by #210

PR #153, opened 2022-07-23
add(package): add const for external reference

masahiro331:
When using this library, we want to use the values defined in the specification.
It would be nice if publish these consts as an official library.

### DIFF
--- a/spdx/package.go
+++ b/spdx/package.go
@@ -8,6 +8,24 @@ import (
 	"strings"
 )
 
+const (
+	CategorySecurity       = "SECURITY"
+	CategoryPackageManager = "PACKAGE-MANAGER"
+	CategoryOther          = "OTHER"
+	CategoryPersistentId   = "PERSISTENT-ID"
+
+	TypeSecurityCPE22Type = "cpe22Type"
+	TypeSecurityCPE23Type = "cpe23Type"
+
+	TypePackageManagerMavenCentral = "maven-central"
+	TypePackageManagerNpm          = "npm"
+	TypePackageManagerNuGet        = "nuget"
+	TypePackageManagerBower        = "bower"
+	TypePackageManagerPURL         = "purl"
+
+	TypePersistentIdSwh = "swh"
+)
+
 type Supplier struct {
 	// can be "NOASSERTION"
 	Supplier string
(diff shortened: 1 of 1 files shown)


swinslow:
Hi @masahiro331, thanks for this!

I'm generally agreed with adding more constants like this. That said, #146 is going to be making extensive changes to the repo, including this file. I'm expecting to merge #146 by next Wednesday if there aren't any objections. 

So after that PR is merged, it will likely make sense to then add these constants in the new `spdx/common/` directory -- maybe in a new file like `spdx/common/constants.go` or something like that.

I'll leave this open so we can track it, but I think I'd suggest doing it in that order so that this one doesn't create a new conflict for #146. Thanks!

lumjjb:
@masahiro331 PR #146 has been merged now... Can you help rebase the code.. I think this should now be part of:

https://github.com/spdx/tools-golang/tree/main/spdx/common

masahiro331:
Thank you,  I have just started this task.

lumjjb:
Awesome - i've assigned the issue to you! Thanks @masahiro331

lumjjb:
Hi @masahiro331 , We have a file for this now : https://github.com/spdx/tools-golang/blob/main/spdx/common/external.go. I'd recommend branching main again and applying your changes to this file!

lumjjb:
Related https://github.com/spdx/tools-golang/pull/163, still looks like the high level categories still need some constants i.e. "PACKAGE-MANAGER"

kzantow:
Hi @masahiro331 -- there was a significant refactoring in this library, are you able to adjust this PR for the changes?

neilnaveen:
> Hi @masahiro331 -- there was a significant refactoring in this library, are you able to adjust this PR for the changes?

I would like to work on this.

neilnaveen:
Should [this file](https://github.com/spdx/tools-golang/blob/main/spdx/v2/common/external.go) be moved back into `spdx/common` and out of `spdx/v2/common`?

Should [the file](https://github.com/spdx/tools-golang/blob/main/spdx/v2/common/external.go) be moved back into the SPDX common folder?

kzantow:
@neilnaveen I don't believe so -- these are already available with aliases at the top-level `github.com/spdx/tools-golang/spdx` package. If it turns out that these are identical or a subset of what `V3` has, I'd say we could move it at that point.

neilnaveen:
Fixed in #188

kzantow:
Hi @masahiro331 thank you for this contribution -- it looks like most of these constants were added as part of #188 so I'm going to close this for now but please reopen if you would like some additional changes!

PR #170, opened 2022-11-07
fix: JSON encoding and decoding not properly handling SPDXRef- prefixes

kzantow:
This PR corrects an issue with JSON encoding & decoding of `ElementID` which was not consistently handling `SPDXRef-` prefixes in IDs -- it was only encoding and decoding these for `DocElementID`. In addition to fixing this inconsistency in the JSON, this makes handling consistent with what [the tag-value parser is doing](https://github.com/spdx/tools-golang/blob/main/tvloader/parser2v3/util.go#L108).

### DIFF
--- a/spdx/common/identifier.go
+++ b/spdx/common/identifier.go
@@ -8,11 +8,57 @@ import (
 	"strings"
 )
 
+const (
+	spdxRefPrefix     = "SPDXRef-"
+	documentRefPrefix = "DocumentRef-"
+)
+
 // ElementID represents the identifier string portion of an SPDX element
 // identifier. DocElementID should be used for any attributes which can
 // contain identifiers defined in a different SPDX document.
 // ElementIDs should NOT contain the mandatory 'SPDXRef-' portion.
 type ElementID string
+
+// MarshalJSON returns an SPDXRef- prefixed JSON string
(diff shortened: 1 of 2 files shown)


spdx/common/identifier.go
```diff
@@ -8,12 +8,58 @@ import (
...
+// UnmarshalJSON validates SPDXRef- prefixes and removes them when processing ElementIDs
+func (d *ElementID) UnmarshalJSON(data []byte) error {
+	// SPDX identifier will simply be a string
+	idStr := string(data)
+	idStr = strings.Trim(idStr, "\"")
+
+	e, err := trimElementIdPrefix(idStr)
+	if err != nil {
+		return err
+	}
```
kzantow, line 36: NOTE: this will mean parsing invalid SPDX files will fail due to invalid SPDX identifiers; unsure if that should be the behavior here. It is consistent with the original `DocElementID.UnmarshalJSON` behavior: https://github.com/spdx/tools-golang/pull/170/files#diff-c46b7836ea8c59261500de448b948294dc3d818d706da8b08dbed529bd1a9609L71

lumjjb: i would prefer to be stricter here, so this should be fine.

lumjjb (approved):
lgtm

Issue #167, opened 2022-10-13
Reduce utility duplication by implementing document conversions


kzantow:
As part of the [suggested refactorings](https://github.com/spdx/tools-golang/issues/165) I proposed, it was noted that the utility functions such as `idsearcher.go`, which duplicate functionality for SPDX 2.1, 2.2, etc. could be made more generic using some reflection (and/or generics, which I don't think will help here).

Additionally, there is a desire to have the ability to convert between format versions. I propose using a "migration" approach for this and have created some functionality that does quite a lot of the work automatically, with heavy use of reflection.

I've created a draft PR here: https://github.com/kzantow-anchore/tools-golang/pull/2 to illustrate how this can work, along with a full document conversion from a 2_2 to a 2_3 struct.

The gist of this approach is twofold:

1. Use reflection to recursively map all fields from one struct to another based on convertible fields:
  - fields with the same name and convertible types are automatically handled
  - slices, maps, structs, and primitive types are currently handled
  - each struct may implement a `ConvertFrom` method, which runs _after_ the automatic conversion to correct, add, or modify the resulting struct, being passed the struct from which it was originally converted
2. Set up a migration list so each version (v2_2, v2_3) can implement a `ConvertFrom` method from the previous (and possibly next) version. Implement a function that, given a specific version and a specific target, walks the list through each intermediate version conversion.

A `ConvertFrom` function for `v2_3.Document` would look something like:
```go
func (t *Document) ConvertFrom(in interface{}) error {
	if prior, ok := in.(v2_2.Document); ok {
           // do any specific stuff here
        }
        t.SPDXVersion = "SPDX-2.3"
	return nil
}
```

Issue #165, opened 2022-10-10
Refactoring based on versions


kzantow:
I've implemented 2.3 support [with this PR](https://github.com/spdx/tools-golang/pull/164). This was not a particularly easy task, as there are a number of inconsistencies in the project layout. It is also not easy to see the differences between the 2.2 and 2.3 code paths. I have a suggestion to refactor this repository such that the layout uses versions at the top level, e.g. something like:

```
common
v2/v2_1
v2/v2_2
v2/v2_3
v2/v2_3/idsearcher/...
v2/v2_3/json/reader.go (with a `Read` function)
v2/v2_3/json/writer.go (with a `Write` function)
v2/v2_3/rdf/reader.go (with a `Read` function)
```

Where each version contains all the version specific files. This avoids files growing uncontrollably like [the idsearcher](https://github.com/spdx/tools-golang/blob/77794935eaa9bc95a1477b73ffb096f0986c1c84/idsearcher/idsearcher.go) as every version increases (there are a number of files like this). It's also unclear when implementing a new version exactly which files need to be modified vs. copied.

There are also inconsistencies like the `json` package containing a `writer` and `parser` but separate packages for `tvloader` and `tvsaver`, there should be a common pattern here (I think the `json` package having a reader and writer is the better pattern). If things are split up by version, then this becomes even easier, as you just reference `<version>/<format>` and have access to a read and write functions and they don't even have to be renamed (just `Read`/`Parse` and `Write`), so these function names with `2_2` and `2_3` aren't things that show up in the diffs between the versions.

If all the code specific to a version was within a specific path, adding a new version would be as simple as copying the previous one and making appropriate changes. It would then be easy to see the differences between them using something like: `diff v2/v2_2 v2/v2_3 --ignore-matching-lines='package v2_.'`

This means for any consumer, all they would need to do is update the paths, like from something like `v2_2` to `v2_3`, the other function calls would remain the same.

I would be happy to take on this refactoring if it is something that might be accepted.

kzantow:
I've made a draft PR to illustrate these changes here: https://github.com/kzantow-anchore/tools-golang/pull/1

I would not be opposed to removing the `v2` subdirectory, so things like 2.3 are `v2_3` instead of `v2/v2_3`, if that is something desired.

This PR is also incomplete:
* many of the comments have erroneous things that should be removed to make transition to a new spec easier
[...]

Issue #160, opened 2022-08-26
Upgrading and downgrading documents


lumjjb:
From: #128 ... Should be partially addressed by https://github.com/spdx/tools-golang/issues/152

* *Upgrading and downgrading documents*: It could be useful to have a function that e.g. takes an SPDX-2.1 document (`spdx.Document2_1`) and converts it to an SPDX_2.2 document (`spdx.Document2_2`). Or vice versa, going the other direction. 
  * This would likely require converting the various sub-elements to the new version.
  * Could also take a `strict` parameter to fail if the conversion can't be done with 100% fidelity. For example, if an SPDX-2.2 document has a field or value that would be invalid in SPDX-2.1, then if `strict` is true then the downgrade from 2.2 to 2.1 should fail. If `strict is false`, then the downgrade would silently (or with a warning) drop the invalid data.

Issue #155, opened 2022-08-26
[WIP] Golang tools roadmap


lumjjb:
2022 2H (v0.4.0)
- [ ] Core model
  - [ ] Mapping from 2.1/2 to core
  - [ ] Do deduplication of code
- [ ] 2.3 support
  - [ ] struct
  - [ ] functions
- [ ] Higher level helper function/libraries to use ([reference bom tool](https://github.com/spdx/tools-golang/issues/144), etc.)
  - [ ] base: here's how to interact with SPDX elements
  - [ ] context-specific: to work with container images, rpm, deb (will use base library)
  - [ ] Upstream use of tools-golang to bom tool
- [ ] Formats
  - [ ] rdfloader - RDF document saver
- [ ] Collect usage patterns
  - [ ] kubernetes-sig/bom
  - [ ] anchore/syft
  - [ ] aquasecurity/trivy
  - [ ] [opensbom-generator/spdx-sbom-generator](https://github.com/opensbom-generator/spdx-sbom-generator)
  - [ ] Organizations
    - [ ] Google (@lumjjb)
    - [ ] Chaingaurd (Adolfo)
    - [ ] VMWare (Rose)
    - [ ] Oracle (Nisha)
- [ ] Misc
  - [ ] https://github.com/spdx/tools-golang/pull/139
  - [ ] https://github.com/spdx/tools-golang/pull/153/
  - [ ] Expand tests for other OS's: see https://github.com/spdx/tools-golang/issues/117
  - [ ] OpenSSF Best Practices badge

2023+
- [ ] SPDX 3.0
- [ ] SBOMs for releases: see https://github.com/spdx/tools-golang/issues/118
- [ ] Collect usage patterns
  - [ ] License usage (@swinslow) - people may want to be able to reason about it as tree (ANDs of licenses), etc. Moving from simple strings to structured data. Figure out how to parse license grammar (pretty involved).
- [ ] Formats
  - [ ] xlsx

lumjjb:
@swinslow @RishabhBhatnagar re-organized this to more realistic timelines and to prioritize a q4 release.

ded1976:
> 2022 2H (v0.4.0)
> - [x] 2.3 support https://github.com/spdx/tools-golang/issues/156
>   - [x] struct
>   - [x] functions
> - [x] Misc
>   - [x] https://github.com/spdx/tools-golang/issues/139
>   - [x] https://github.com/spdx/tools-golang/issues/153 
> 
> 
[...]

PR #172, opened 2022-11-22
refactor!: maintain the latest spdx model and provide conversions from previous

kzantow:
This PR implements a fairly straightforward way to maintain a data model all utilities and consumers are able to use.

The gist of this approach is:
* define all data models and directly related functions under `spdx` (`spdx/v2_1`, `spdx/v2_2`, etc.)
* maintain _type aliases_ to [the latest/current data model](https://github.com/spdx/tools-golang/pull/172/files#diff-0ded6e02d307b3c4c175d93da5bc396716c8abba2b8d93b7af34159a2706876b)
* provide migrations between versions
* all top-level functions only operate on the latest/current model
  * e.g. utilities in `builder`, `examples`, `idsearcher`, `licensediff`, `reporter`, `spdxlib`, `utils` all operate on `spdx.Document` and do not have duplication for older formats
* all top-level format `Read` functions always convert to the latest/current data model:
  * e.g. `tagvalue.Read(<2.1 document>)` reads a 2.1 document, converts to and returns the current `spdx.Document`
  * all format types _also_ provide a `ReadInto`, however, which converts directly to the provided doc type
* align naming of all readers/writers to use terms `Read` and `Write`
* remove all function naming with version specifics, instead refactor these to the corresponding `spdx/<version>` packages
* make imports as consistent as possible, including aliasing all version-specific model imports to `spdx`
  * this makes a copy/paste operation of a new version significantly easier as basically _only imports_ need to be updated

Ultimately, this change should allow downstream consumers to utilize this library in a significantly easier manner, as an example:

1) Consumer incorporates this library to read SPDX files, using some code similar to:
```go
import "github.com/spdx/tools-golang/json"

func SomeFunction() {
  doc, err := json.Read(something)
  // deal with doc, which is `spdx.Document`, currently at v2.3
}
```

2) New SPDX version is released (let's say `2.4`), `tools-golang` is updated. Type aliases now point to `2.4` instead of `2.3`. E.g. `spdx.Document` -> `v2_4.Document`.

3) Consumer updates library and wants `2.4` support, the code before _doesn't need to change_, and automatically handles reading older 2.3 SPDX files. The only thing the consumer needs to account for is changes in the data model, which they would have to do in all cases to support a newer SPDX version.

NOTE: I have updated [Syft](https://github.com/anchore/syft) to use this model in a working branch -- see: https://github.com/anchore/syft/compare/main...kzantow-anchore:syft:chore/refactored-spdx-lib?expand=1

TODO:
- [x] ~handle tag-value conversions more generically to reduce per-version duplication~
- [x] ~handle rdf decoding more generically to reduce per-version duplication~
- [x] update documentation

Fixes: #175 
Fixes: #167
Fixes: #165
Fixes: #160


convert/struct.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/spdx/tools-golang/spdx/common"
+)
+
+// FromPtr accepts a document or a document pointer and returns the direct struct reference
+func FromPtr(doc common.Document) common.Document {
```
lumjjb, line 11: does `FromPtr` and `Describe` have to be public?

json/writer.go
```diff
@@ -1,32 +1,16 @@
...
-		return err
-	}
-
-	return nil
-}
-
-// Save2_3 takes an SPDX Document (version 2.2) and an io.Writer, and writes the document to the writer in JSON format.
-func Save2_3(doc *v2_3.Document, w io.Writer) error {
+// Write takes an SPDX Document and an io.Writer, and writes the document to the writer in JSON format.
+func Write(doc common.Document, w io.Writer) error {
```
lumjjb, line 13: Curious about the chain converter you've implemented, what would happen if we created another chain for backwards conversion and have an option to say what version it would like to be exported as?

lumjjb (commented):
@kzantow this is a humungous effort and this is truly a 1000x improvement for the golang library and SPDX more broadly!!! The proposed changes look good, and I've added some minor comments/questions.. 

The only part of it that i'd want to discuss more about is with v3, it is going to be a breaking change for consumers of the library. With your changes here, when we upgrade from 2.2 to 2.3, it is truly just a package update - which is SUPER SUPER SUPER awesome!! However, this is likely impossible with v3 - since it has a totally different data model and fields, so i'm wondering if it's maybe still better to put it into a folder as `spdx/v2` or just `spdxv2/` as the base package name. Thoughts?

convert/struct.go
kzantow: Yes, they're used by different packages

json/writer.go
kzantow: It already supports both forward and backwards conversion!

Let's say you have the latest model and want to export v2.2, the code is something like this:

```go
// doc is an spdx.Document from somewhere, or it could be a different version, like v2_2.Document etc.
doc := spdx.Document{}
// make the result document version we want:
out := v2_2.Document{}
// call convert, which handles the backwards migration as well:
convert.DocumentChain.Convert(doc, &out)
```

kzantow:
@lumjjb thanks much for the review -- regarding v3, I would very much like to know more about this, I'll ping you about it!

README.md
```diff
@@ -6,11 +6,15 @@
 
 # SPDX tools-golang
 
-tools-golang is a collection of Go packages intended to make it easier for
+`tools-golang` is a collection of Go packages intended to make it easier for
 Go programs to work with [SPDX®](https://spdx.dev/) files.
 
 ## Recent news
 
+2023-01: **v0.5.0**: refactored the data model to be consistent for consumers
```
kzantow, line 14: Added this version based on the roadmap here: https://github.com/spdx/tools-golang/issues/155

README.md
```diff
@@ -23,15 +27,14 @@ of the SPDX specification, available at: https://spdx.dev/specifications
 tools-golang provides the following packages:
 
 * *spdx* - in-memory data model for the sections of an SPDX document
-* *tvloader* - tag-value document loader
-* *tvsaver* - tag-value document saver
-* *rdfloader* - RDF document loader
-* *json* - JSON document parser and writer
-* *yaml* - YAML document parser and writer
+* *tagvalue* - tag-value document reader and writer
```
kzantow, line 30: Is there a preference to use `tagvalue` vs. `tv`?

ivanayov:
Thank you for addressing this @kzantow ! I've been struggling with subversion upgrades and am looking forward to adopting this!
 
My only comment is that currently it always defaults to the latest SPDX, while one may want to configure a specific version. For example, v2.2 is ISO approved and v2.3 - not. For that reason some people and organisations use only v2.2 prior the next major version. But others may prefer the latest 2.3. It would be great to allow freedom of configuring which version to default to. 

This would make the v3.0 transition easier as well.

Other than that, I very much like how you approach it.

kzantow:
Thanks @ivanayov -- I'm actually almost done on a last change to make using specific versions easier. I'll update with some examples shortly.

kzantow:
@ivanayov I'll write up some usage examples in the documentation after this PR gets merged, but to answer your question about supporting a _specific version_, I'd suggest this approach:

First: always _alias your imports_ of a specific data model as `spdx`

For example:

```go
import (
  "github.com/spdx/tools-golang/spdx/json"
  spdx "github.com/spdx/tools-golang/spdx/v2_2"
)

// you will always get the specific version you are looking for, assuming a conversion is possible:
var doc spdx.Document
json.ReadInto(<reader>, &doc) // reads _any version_ and converts it to a 2.2 version

// all the write functions work with the specific version you pass in:
json.Write(&doc) // writes a 2.2 version
```

This makes upgrading _very simple_ by just updating _imports_ instead of every single usage of the types. If you _always want to be at the latest version_, then the import is just `"github.com/spdx/tools-golang/spdx"`.

ivanayov:
I think that works well. Thanks @kzantow !

README.md
lumjjb: `tagvalue` is fine!

convert/spdx_document_conversion_test.go
```diff
@@ -0,0 +1,2133 @@
...
+	}
+}
+
+func toJSON(data interface{}) string {
+	bytes, err := json.Marshal(data)
+	if err != nil {
+		panic(err)
+	}
+	return string(bytes)
+}
```
lumjjb, line 2133: these tests are tremendously helpful!

convert/chain.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/spdx/tools-golang/spdx/v2/v2_3"
+)
+
+var DocumentChain = converter.NewChain(
+	v2_1.Document{},
+	v2_2.Document{},
+	v2_3.Document{},
+)
+
+// Document converts from one document to another document
```
lumjjb, line 23: nit: add a couple words here to provide usage may be helpful since it may be tricky to figure out what the arguments look like without reading the code (especially that its expecting a pointer)

README.md
```diff
@@ -6,7 +6,7 @@
 
 # SPDX tools-golang
 
-tools-golang is a collection of Go packages intended to make it easier for
+`tools-golang` is a collection of Go packages intended to make it easier for
```
lumjjb, line 9: I think it will be super helpful to link your usage repo somewhere in this README or as a comment in the root `/spdx` package as an example of how to use this new refactored version!

json/writer.go
```diff
@@ -1,32 +1,16 @@
...
-		return err
-	}
-
-	return nil
-}
-
-// Save2_3 takes an SPDX Document (version 2.2) and an io.Writer, and writes the document to the writer in JSON format.
-func Save2_3(doc *v2_3.Document, w io.Writer) error {
+// Write takes an SPDX Document and an io.Writer, and writes the document to the writer in JSON format.
+func Write(doc common.AnyDocument, w io.Writer) error {
```
lumjjb, line 13: probably a next step is for all output encoders to provide an option of the version requested, we'll open a separate issue for this, not a blocker.

lumjjb (dismissed):
Thank you very much again @kzantow on helping with this very complex and extensive PR! LGTM! I have some minor documentation nits... happy to merge this by end of week if no other comments. @swinslow @RishabhBhatnagar

json/writer.go
kzantow: the `Write` functions accept all versions and output the corresponding file, e.g. a v2_2.Document will generate a 2.2 version if passed in. I'd like this to be more strongly typed. The only way I could think of doing this is to create an interface with an arbitrary function, and implement it for each `Document` types...

README.md
kzantow: The problem with the usage repo is it's just a branch that will probably get deleted. Once this is merged and I've updated Syft, I could have a follow-on to the README that links directly to the Syft usage but probably shouldn't put any "permanent" links before that.

lumjjb (approved):
LGTM!

Issue #156, opened 2022-08-26
SPDX 2.3 Support


lumjjb:
Implement SPDX 2.3 support (preferably waiting on https://github.com/spdx/tools-golang/issues/152) for refactor changes. 

2.3 Link: https://spdx.github.io/spdx-spec/v2.3/

nishakm:
Need help with this? I need a 2.3 model to implement a generator :)

lumjjb:
that would be super helpful!!!!

kzantow:
@nishakm I also need some of the new features of 2.3 -- could I help with this?

nishakm:
> @nishakm I also need some of the new features of 2.3 -- could I help with this?

Absolutely!

lumjjb:
Assigned the issue to bot of you!!! Thank you!!

kzantow:
I'm starting to look at implementing this -- is there a document somewhere with the changes for 2.3?

lumjjb:
Awesome @kzantow !!!! Yes! Here's the changes from 2.2 (https://spdx.github.io/spdx-spec/v2.3/diffs-from-previous-editions/)! Is this something that's useful?

You can ignore the between v2.2.x versions because they are mainly documentation changes.

kzantow:
@lumjjb I've [added a draft PR](https://github.com/spdx/tools-golang/pull/164) with some initial changes for 2.3: namely a new v2_3 data model and JSON support, along with a number of other changes. It does not yet have the other format support (it will before I'd consider it done).

It looks like you did some of the refactoring for the current state of this repo, if you had a little time, could you have a quick look and make sure I'm on the right track before I complete everything else?

kzantow:
@lumjjb I believe I've implemented all the 2.3 support in this PR: https://github.com/spdx/tools-golang/pull/164

kzantow:
I've also added a refactoring suggestion I would like to do here: https://github.com/spdx/tools-golang/issues/165 I know there are some other things (like translating between versions) that people would like to see, but I feel that's a separate issue and the refactoring would dramatically improve the experience every time someone needs to implement a new SPDX version.

PR #174, opened 2022-12-05
Ensure consistency between JSON struct tags across different SPDX versions

jedevc:
Related to #156.

It looks like some of the struct tags between different SPDX versions (v2.1, v2.2 and v2.3) were slightly out-of-sync. This results in some of the deprecated fields, non-spec compliant fields being exported by the `json.Save` family of functions, but inconsistently between versions. Additionally, some optional comment fields were not marked with `omitempty`, resulting in unnecessary JSON output.

This PR fixes these various inconsistencies.

### DIFF
--- a/spdx/v2_1/creation_info.go
+++ b/spdx/v2_1/creation_info.go
@@ -22,5 +22,5 @@ type CreationInfo struct {
 
 	// 2.10: Creator Comment
 	// Cardinality: optional, one
-	CreatorComment string `json:"comment"`
+	CreatorComment string `json:"comment,omitempty"`
 }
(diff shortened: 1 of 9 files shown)


kzantow:
Unfortunately, I'm not a maintainer but the changes LGTM 👍

thaJeztah:
Thanks for looking @kzantow ! Reviews from non-maintainers never hurt; can always use the extra eyes 👀

lumjjb (approved):
LGTM! Sorry was out on vacation past few weeks! Thanks for the PR and apologies for the delayed response.

Issue #166, opened 2022-10-12
Update to add `documentDescribes` field to follow 2.2.2 spec


lucacome:
I noticed that there are some fields missing, like `documentDescribes`, that I think were introduced with version `2.2.1` of the spec.

It'd be great if all of the fields in the https://github.com/spdx/spdx-spec/blob/v2.2.2/schemas/spdx-schema.json could be added.

Thanks! 🙏

lumjjb:
Thanks @lucacome for opening this issue - sorry for the late reponse. We will take a look at this and have it be part of the 0.4.0 milestone

lucacome:
Thanks @lumjjb, it looks like also `2.3` came out in the meantime https://github.com/spdx/spdx-spec/blob/v2.3/schemas/spdx-schema.json

lumjjb:
It seems like one way to implement this is through the json `Save` function and the addition of the `documentDescribes` field in the Document struct. Let me write up a PR for this.

lumjjb:
There looks like there's some weirdness in mismatch between spec and example for 2.2.0, will clarify this with the SPDX tech list first before implementing.

lucacome:
I think this can be closed now?

lumjjb:
This issue is not quite fully resolved yet. I had a conversation with @kestewart and we probably need to surface this to the tech working group to get some consensus on these additional fields that are not official spec fields or to be able to formalize the JSON schema spec. I will send something to the mailing list and hope that will bring a conclusion to the issue?

lumjjb:
>I just checked and it looks like “documentDescribed” is already optional in the JSON Schema, which should relieve some of the pressure on this issue.

So actually we don't have to implement it...... technically.... so yes let's close the issue for now and re-open if something arises.. Thanks for checking back @lucacome !

goneall:
Even though it is optional, it may be useful to be able to read (no necessarily to write) this field since some of the other language bindings support this field.

lumjjb:
Opening this ticket again since the current conversation on the mailing list looks towards formalizing this field.

lumjjb:
copying link of email chain here: https://lists.spdx.org/g/Spdx-tech/topic/96047024

PR #176, opened 2022-12-20
feat: add required document describes field

spiffcs:
While it's not described in the documentation, the `DocumentDescribes` field can be found in the schema here:
https://github.com/spdx/spdx-spec/blob/efa69656db9db4fcc871de563cbbd104fc1e33c3/schemas/spdx-schema.json#L219-L226

As of this commit it's considered a required field. Tools like the java-tools, python-tools, and online validator will prompt the user that documents generated without this field are invalid:

https://github.com/spdx/spdx-spec/commit/efa69656db9db4fcc871de563cbbd104fc1e33c3

Adding this field for the v2_3 document will allow tools like [syft](https://github.com/anchore/syft) which use this library to produce valid documents.

Thanks a million for all the work and effort on this library!

Fixes: #166 

Signed-off-by: Christopher Phillips <christopher.phillips@anchore.com>

### DIFF
--- a/spdx/v2_3/document.go
+++ b/spdx/v2_3/document.go
@@ -44,6 +44,11 @@ type Document struct {
 	// Cardinality: mandatory, one
 	DocumentNamespace string `json:"documentNamespace"`
 
+	// 6.x: Document Descibes
+	// Cardinality: mandatory
+	// Packages, files and/or Snippets described by this SPDX document
+	DocumentDescribes []string `json:"documentDescribes"`
+
 	// 6.6: External Document References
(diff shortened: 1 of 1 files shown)


spiffcs:
Closed - @kzantow informed me that this is a "shorthand" property that is a duplicate of a relationship used to accomplish the same thing. 

In this case, without these properties, I think we just need a relationship and the validators should be valid (see: [here](https://spdx.github.io/spdx-spec/v2.3/relationships-between-SPDX-elements/)). The relationship needs to be type = DESCRIBES with the to = SPDXRef-DOCUMENT

PR #180, opened 2023-01-11
docs: Remove outdated statement about requirements

swinslow:
The README file currently states that tools-golang has no dependencies other than spdx/gordf and the Go standard library.

With the addition of the YAML functionality and usage of go-cmp, this is no longer an accurate statement. Rather than calling out the specific dependencies, this commit just removes that section. Users are able to check go.mod and go.sum if they want to know which dependencies are required.


lumjjb (approved):
lgtm

Issue #182, opened 2023-01-17
tagvalue parser incorrectly complains about DocumentComment


justinabrahms:
Hello.

When parsing a tagvalue file, I incorrectly get the error `received unknown tag DocumentComment in CreationInfo section`. I've ensured the file passes validation at https://tools.spdx.org/app/validate/.

File listed here: https://github.com/eBay/sbom-scorecard/issues/22#issuecomment-1381069619

Test case:

```go
package main

import (
	"bytes"
	"fmt"
	"io/ioutil"

	spdx_tv "github.com/spdx/tools-golang/tvloader"
)

func main() {
	f, err := ioutil.ReadFile("/Users/jabrahms/Downloads/photon.spdxtv.txt")
	if err != nil {
		print("error opening SPDX document: %w", err)
		return
	}

	doc22, err := spdx_tv.Load2_2(bytes.NewReader(f))
	if err == nil && doc22 != nil {
		print("Found 2.2 tv document")
	} else {
		fmt.Println("Error: ", err)
	}
}
```

neilnaveen:
@lumjjb  I do not see tvloader package in main. It was their in the release v0.4.0, but now it is gone, how do you want to handle this?

justinabrahms:
@neilnaveen I'm happy if you just fix what's in main.

```go
package main

import (
	"bytes"
	"fmt"
	"io/ioutil"

	spdx_tv "github.com/spdx/tools-golang/tagvalue"
)

func main() {
	f, err := ioutil.ReadFile("/Users/jabrahms/Downloads/photon.spdxtv.txt")
	if err != nil {
		print("error opening SPDX document: %w", err)
		return
	}

	doc22, err := spdx_tv.Read(bytes.NewReader(f))
	if err == nil && doc22 != nil {
		print("Found 2.2 tv document")
	} else {
		fmt.Println("Error: ", err)
	}
}
```

and
```
$ go run main.go
Error:  received unknown tag DocumentComment in CreationInfo section
```

neilnaveen:
Not in your code, but the [main branch](https://github.com/spdx/tools-golang) of spdx.

justinabrahms:
@neilnaveen Totally. I (I think...?) updated the code to reflect the test case with what's in main.

lumjjb:
@justinabrahms does the PR https://github.com/spdx/tools-golang/pull/185 from @neilnaveen fix this for you?

lumjjb:
[...]

PR #185, opened 2023-01-26
Fixed Bug For DocumentComment

neilnaveen:
addresses issue #182 

- added DocumentComment switch case for tag in parse_creation_info.go
- added test for DocumentComment switch case
- added file parse_document.go for the switch case above
- added test for parse_document.go

Signed-off-by: Neil Naveen <42328488+neilnaveen@users.noreply.github.com>


lumjjb (approved):
LGTM! Thanks @neilnaveen

spdx/v2/v2_2/tagvalue/reader/parse_creation_info_test.go
```diff
@@ -78,6 +78,18 @@ func TestParserCIMovesToOtherLicenseAfterParsingLicenseIDTag(t *testing.T) {
 	}
 }
 
+func TestParserCIMovesToOtherLicenseAfterParsingDocumentTag(t *testing.T) {
```
kzantow, line 81: I think this name might be a little misleading, should it be `TestDocumentComment` or something more directly related to that?

kzantow (commented):
This change should also be applied to `spdx/v2/v2_3/tagvalue/...` files

spdx/v2/v2_2/tagvalue/reader/parse_creation_info_test.go
neilnaveen: I agree, typo, will fix in the following PR.

PR #188, opened 2023-02-04
Added more const for external reference to external.go

neilnaveen:
- Added more const for external.go, from the PR https://github.com/spdx/tools-golang/pull/153

Signed-off-by: Neil Naveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/spdx/v2/common/external.go
+++ b/spdx/v2/common/external.go
@@ -4,6 +4,10 @@ package common
 
 // Constants for various string types
 const (
+	CategorySecurity       = "SECURITY"
+	CategoryPackageManager = "PACKAGE-MANAGER"
+	TypePersistentIdSwh    = "swh"
+
 	// F.2 Security types
 	TypeSecurityCPE23Type string = "cpe23Type"
 	TypeSecurityCPE22Type string = "cpe22Type"


spdx/v2/common/external.go
```diff
@@ -4,6 +4,10 @@ package common
 
 // Constants for various string types
 const (
+	CategorySecurity       = "SECURITY"
+	CategoryPackageManager = "PACKAGE-MANAGER"
+	TypePersistentIdSwh    = "swh"
```
lumjjb, line 9: looks good - can we create a comment to show which section of the spec they come from? (like below). After this PR would you be able to help with a follow-up PR to change all uses of the categories to use the const as well? :D

lumjjb (commented):
Thanks @neilnaveen for opening this up!

spdx/v2/common/external.go
neilnaveen: What section do you think `TypePersistentIdSwh` should go into

kzantow: I think the Categories are part of [external references](https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field). `swh` is from [Annex F](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f4-persistent-id).

kzantow (commented):
Hey @neilnaveen if you are going to add these constants (which seems good 👍), could you add _all_ the constants defined in the external references categories and [annex f types](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/)?

Also these need [aliases in the spdx/model.go](https://github.com/spdx/tools-golang/blob/main/spdx/model.go#L65).

I'd second @lumjjb 's request to replace the hardcoded values with these constants where appropriate, too. I think it would make sense to be part of this PR.

spdx/v2/common/external.go
neilnaveen: Thanks, I have fixed this

neilnaveen:
> Hey @neilnaveen if you are going to add these constants (which seems good 👍), could you add _all_ the constants defined in the external references categories and [annex f types](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/)?
> 
> Also these need [aliases in the spdx/model.go](https://github.com/spdx/tools-golang/blob/main/spdx/model.go#L65).
> 
> I'd second @lumjjb 's request to replace the hardcoded values with these constants where appropriate, too. I think it would make sense to be part of this PR.

I have added all constants from the annex f types, and I have put them in their respective categories.

kzantow (commented):
Hi @neilnaveen this PR stills seems to be missing a number of changes from the original PR: https://github.com/spdx/tools-golang/pull/153/files it also does not include the aliases in the `spdx` package, are you able to fix this?

neilnaveen:
> Hi @neilnaveen this PR stills seems to be missing a number of changes from the original PR: https://github.com/spdx/tools-golang/pull/153/files it also does not include the aliases in the `spdx` package, are you able to fix this?

I have added aliases into spdx/model.go. Could you also tell me which changes I have not added in this PR that was in #153?

kzantow (approved):

PR #197, opened 2023-02-23
implement documentDescribes for json/yaml schema

lumjjb:
Addressing https://github.com/spdx/tools-golang/issues/166
- Update json document examples for 2.2 and 2.3 to include documentDescribes as updated in upstream schema.
- Add field DocumentDescribesJSON field to document struct for 2.2 and 2.3 to match JSON schema
- Implement populating of documentDescribes field based on relationships
- Added TODO to do this for all similar fields like documentDescribes

Drive-bys:
- fix small bug in sdpx/v2/v2_2/yaml test


lumjjb:
close in favor of #201

Issue #171, opened 2022-11-08
JSON readers are missing `hasFiles` handling


kzantow:
Part of the SPDX 2.x JSON schema includes [a `hasFiles` field on packages](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json#L345-L352). There is a reference to `hasFiles` in this repo's docs [here](https://github.com/spdx/tools-golang/blob/main/docs/jsonsaver.md?plain=1#L11) and [here](https://github.com/spdx/tools-golang/blob/main/docs/jsonloader.md?plain=1#L11) but there is no field in the model for this (which is probably fine). For reading SPDX 2.x documents, however, this field should not be ignored and should probably be mapped to `CONTAINS` relationships.

More information here: https://github.com/spdx/spdx-spec/issues/805#issuecomment-1306459467

kzantow:
@lumjjb as I understand it (and as it was explained in the [response](https://github.com/spdx/spdx-spec/issues/805#issuecomment-1306459467)) the `hasFiles` has been deprecated since 2.0 and is equivalent to relationship `CONTAINS`. I would suggest _not_ adding the field so users of the library generate SBOMs using the `CONTAINS` method only.

But to support SBOMs that are generated from other tools, it seems like it is important to _read_ this field. However, since it is equivalent to the `CONTAINS` relationship, I was suggesting instead of adding `hasFiles`, there could be some custom JSON unmarshal step that reads this but converts it to a `CONTAINS` relationship.

What do you think?

goneall:
Just an FYI - If I remember correctly, the Java tools converts `hasFiles` to a `CONTAINS` relationship for the RDF/XML format but retains the `hasFiles` property.  A bit inconsistent, but I think there was a request to retain the `hasFile` for JSON readability.  For the 3.0 version I'm hoping to clean this up and make things much more consistent.

PR #201, opened 2023-03-04
fix: read shortcut fields: documentDescribes and hasFiles

kzantow:
The JSON schema has fields [documentDescribes](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json#L219) and [hasFiles](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json#L345) for packages. These are not part of the Go data model and are duplicates of using relationships. However, some tools are outputting these fields today and we need to be able to decode them. This PR adds support for decoding these fields and translating them into the appropriate relationships.

Fixes: #166 
Fixes: #171

### DIFF
--- a/spdx/v2/v2_2/package.go
+++ b/spdx/v2/v2_2/package.go
@@ -3,6 +3,8 @@
 package v2_2
 
 import (
+	"encoding/json"
+
 	"github.com/spdx/tools-golang/spdx/v2/common"
 )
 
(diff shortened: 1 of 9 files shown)


spdx/v2/v2_2/package.go
```diff
@@ -113,8 +115,37 @@ type Package struct {
...
+
+	// this field is only used when decoding JSON to translate the hasFiles
+	// property to relationships
+	hasFiles []common.DocElementID
+}
+
+func (p *Package) UnmarshalJSON(b []byte) error {
+	type pkg Package
+	type extras struct {
+		HasFiles []common.DocElementID `json:"hasFiles"`
```
lumjjb, line 127: ahhh i see this trick to pass the `hasFiles` up to unexposed fields, this is very cool :)

lumjjb (approved):
super cool implementation! lgtm

puerco:
This is great! Thanks @kzantow :tada:

PR #203, opened 2023-03-24
feat: normalize external reference category when reading JSON and YAML

### LINKED ISSUE #196: SPDX JSON v2.2 should output PACKAGE_MANAGER
The [SPDX 2.2 JSON schema](https://github.com/spdx/spdx-spec/blob/development/v2.2/schemas/spdx-schema.json#L273) only defines `PACKAGE_MANAGER`, not `PACKAGE-MANAGER` as allowed for the external reference category. This is in conflict with the spec and tag-value, [which only defines the latter, `PACKAGE-MANAGER`](https://spdx.github.io/spdx-spec/v2.2.2/package-information/#721-external-reference-field). However, this causes problems with any tools using the JSON schema to validate SPDX documents. Specifically, for SPDX 2.2 JSON should transform this value when writing and reading. As tools are supposed to accept _both_ values, this can help a bit to normalize to `PACKAGE-MANAGER` in the `tools-golang` library.

kzantow:
This PR normalizes the `referenceCategory` for external references when reading JSON and YAML to read the alternate values with underscores, like `PERSISTENT_ID` and `PACKAGE_MANAGER` and translate these [to the values in the SPDX spec](https://spdx.github.io/spdx-spec/v2.3/package-information/#7211-description), which instead have dashes: `PERSISTENT-ID` and `PACKAGE-MANAGER`, respectively. This allows tools reading and writing SPDX to simply use the spec values with dashes.

Fixes #196

### DIFF
--- a/spdx/v2/v2_2/example/example.go
+++ b/spdx/v2/v2_2/example/example.go
@@ -203,7 +203,7 @@ var example = spdx.Document{
 			PackageDownloadLocation: "https://search.maven.org/remotecontent?filepath=org/apache/jena/apache-jena/3.12.0/apache-jena-3.12.0.tar.gz",
 			PackageExternalReferences: []*spdx.PackageExternalReference{
 				{
-					Category: "PACKAGE_MANAGER",
+					Category: "PACKAGE-MANAGER",
 					RefType:  "purl",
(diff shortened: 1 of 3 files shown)


kzantow:
Closing this in favor of: https://github.com/spdx/tools-golang/pull/204

PR #218, opened 2023-05-23
Ensure no duplicates in relationships when shortcut fields are used.

### LINKED ISSUE #208: Check for duplicates when unmarshaling shortcut fields
In PR https://github.com/spdx/tools-golang/pull/201, we introduced the auto unparsing of the shortcut "documentDescribes" and "hasFiles" fields.

It doesn't seem that in the code that this checks for duplicates, and may result in duplicate relationships.

lumjjb:
In the unmarshaling of shorthand fields, checks if relationships exists before adding them.

Fixes #208

### DIFF
--- a/spdx/v2/v2_2/document.go
+++ b/spdx/v2/v2_2/document.go
@@ -5,8 +5,9 @@ package v2_2
 
 import (
 	"encoding/json"
+	"fmt"
 
-	"github.com/anchore/go-struct-converter"
+	converter "github.com/anchore/go-struct-converter"
 
 	"github.com/spdx/tools-golang/spdx/v2/common"
 )
@@ -98,27 +99,46 @@ func (d *Document) UnmarshalJSON(b []byte) error {
 
 	*d = Document(d2)
 
+	relationshipExists := map[string]bool{}
+	serializeRel := func(r *Relationship) string {
+		return fmt.Sprintf("%v-%v->%v", common.RenderDocElementID(r.RefA), r.Relationship, common.RenderDocElementID(r.RefB))
+	}
+
+	// index current list of relationships to ensure no duplication
+	for _, r := range d.Relationships {
+		relationshipExists[serializeRel(r)] = true
+	}
+
 	// build relationships for documentDescribes field
 	for _, id := range e.DocumentDescribes {
-		d.Relationships = append(d.Relationships, &Relationship{
+		r := &Relationship{
 			RefA: common.DocElementID{
 				ElementRefID: d.SPDXIdentifier,
 			},
 			RefB:         id,
 			Relationship: common.TypeRelationshipDescribe,
-		})
+		}
+
+		if !relationshipExists[serializeRel(r)] {
+			d.Relationships = append(d.Relationships, r)
+			relationshipExists[serializeRel(r)] = true
+		}
 	}
 
 	// build relationships for package hasFiles field
 	for _, p := range d.Packages {
 		for _, f := range p.hasFiles {
-			d.Relationships = append(d.Relationships, &Relationship{
+			r := &Relationship{
 				RefA: common.DocElementID{
 					ElementRefID: p.PackageSPDXIdentifier,
 				},
 				RefB:         f,
 				Relationship: common.TypeRelationshipContains,
-			})
+			}
+			if !relationshipExists[serializeRel(r)] {
+				d.Relationships = append(d.Relationships, r)
+				relationshipExists[serializeRel(r)] = true
+			}
 		}
 
 		p.hasFiles = nil
--- a/spdx/v2/v2_2/json/json_test.go
+++ b/spdx/v2/v2_2/json/json_test.go
@@ -4,6 +4,7 @@ package json
 
 import (
 	"bytes"
+	jsonenc "encoding/json"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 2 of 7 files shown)


spdx/v2/v2_2/document.go
```diff
@@ -98,27 +99,46 @@ func (d *Document) UnmarshalJSON(b []byte) error {
...
+	relationshipExists := map[string]bool{}
+	serializeRel := func(r *Relationship) string {
+		return fmt.Sprintf("%v-%v->%v", common.RenderDocElementID(r.RefA), r.Relationship, common.RenderDocElementID(r.RefB))
+	}
+
+	// index current list of relationships to ensure no duplication
+	for _, r := range d.Relationships {
+		relationshipExists[serializeRel(r)] = true
+	}
+
```
kzantow, line 111: What would you say about replacing this with something more like: 
```go
	allRelationships := map[Relationship]interface{}{}

	// index current list of relationships to ensure no duplication
	for _, r := range d.Relationships {
		allRelationships[*r] = nil
	}
```

spdx/v2/v2_2/document.go
```diff
@@ -98,27 +99,46 @@ func (d *Document) UnmarshalJSON(b []byte) error {
...
 			},
 			RefB:         id,
 			Relationship: common.TypeRelationshipDescribe,
-		})
+		}
+
+		if !relationshipExists[serializeRel(r)] {
+			d.Relationships = append(d.Relationships, r)
+			relationshipExists[serializeRel(r)] = true
+		}
```
kzantow, line 125: With the above change (which I've read might use less memory?), this section would be:
```go
		r := Relationship{
			RefA: common.DocElementID{
				ElementRefID: d.SPDXIdentifier,
			},
			RefB:         id,
			Relationship: common.TypeRelationshipDescribe,
		}

		if _, ok := allRelationships[r]; !ok {
			d.Relationships = append(d.Relationships, &r)
			allRelationships[r] = nil
		}
```

kzantow (commented):
Included a minor suggestion for refactoring to possibly use less memory and also avoid creating a bunch of strings.

spdx/v2/v2_2/document.go
kzantow: Ah you may be right about that. Unfortunate! We could refactor the relationship to be a relationship and a relationship with comment, e.g.:
```go
type BasicRelationship struct {
	RefA         common.DocElementID `json:"spdxElementId"`
	RefB         common.DocElementID `json:"relatedSpdxElement"`
	Relationship string
}

type Relationship struct {
	BasicRelationship

	RelationshipComment string `json:"comment,omitempty"`
}
```
... in which case it would be `allRelationships := map[BasicRelationship]struct{}{}`

kzantow: I don't think this would change any API surface area for users, they would just continue to reference Relationship

lumjjb: I think this would change the way they create Relationships though - i think golang doesn't allow you to declare it embedded? but i may be wrong there

kzantow: If you have an embedded struct like:
```go
type Relationship struct {
	BasicRelationship
}
```
The fields (and methods) from BasicRelationship get "promoted" to the top-level  Relationship struct, e.g. I can write:
```go
r := Relationship{}
r.RefA = <something>
```
... and everything works as expected. Additionally, the go JSON library treats embedded fields like this without nesting them in the JSON structure.

kzantow: This definitely won't work as I had thought, so belay all talk of this embedded struct idea, sorry!

spdx/v2/v2_2/json/json_test.go
```diff
@@ -200,7 +172,136 @@ func Test_ShorthandFields(t *testing.T) {
 				Relationship: common.TypeRelationshipContains,
 			},
 		},
-	}, doc)
+	}
+
+	if diff := cmp.Diff(want, doc, cmpopts.IgnoreUnexported(spdx.Package{}), cmpopts.SortSlices(relationshipLess)); len(diff) > 0 {
+		t.Errorf("got incorrect struct after parsing JSON example: %s", cmp.Diff(want, doc, cmpopts.IgnoreUnexported(spdx.Package{})))
```
kzantow, line 178: `cmp.Diff(want, doc, cmpopts.IgnoreUnexported(spdx.Package{}))` should just be `diff` here, right?

spdx/v2/v2_3/json/json_test.go
```diff
@@ -190,7 +191,135 @@ func Test_ShorthandFields(t *testing.T) {
 				Relationship: common.TypeRelationshipContains,
 			},
 		},
-	}, doc)
+	}
+
+	if diff := cmp.Diff(want, doc, cmpopts.IgnoreUnexported(spdx.Package{}), cmpopts.SortSlices(relationshipLess)); len(diff) > 0 {
+		t.Errorf("got incorrect struct after parsing JSON example: %s", cmp.Diff(want, doc, cmpopts.IgnoreUnexported(spdx.Package{})))
```
kzantow, line 197: `cmp.Diff(want, doc, cmpopts.IgnoreUnexported(spdx.Package{}))` should be `diff` here too,  I think

kzantow (approved):
Just a couple things I suspect got overlooked using the `diff` output. Otherwise LGTM! 👍

kzantow (approved):
👍

PR #220, opened 2023-06-06
fix duplicate shorthand relationships for opposite case

### LINKED ISSUE #208: Check for duplicates when unmarshaling shortcut fields

lumjjb:
Really fix #208 :). We forgot to check for the opposite relationships (e.g. CONTAINS_BY and CONTAINED_BY are both considered the same and a new CONTAINS relationship shouldn't be added by shorthand field)


kzantow (approved):
LGTM

PR #221, opened 2023-07-27
Fix: SPDX decode error originator

spiffcs:
## Summary

Given the field `"originator" : "Organization:SPDX",` `tools-golang` would error on decoding a valid SBOM.

This change adds test protection to the reader for any future changes to 2_3 custom unmarshalling as well as fixes the underlying error where the originator field will consider both `: ` and `:` as valid split cases.

### DIFF
--- a/spdx/v2/common/package.go
+++ b/spdx/v2/common/package.go
@@ -70,17 +70,13 @@ func (o *Originator) UnmarshalJSON(data []byte) error {
 		return nil
 	}
 
-	originatorFields := strings.SplitN(originatorStr, ": ", 2)
-
+	originatorFields := strings.SplitN(originatorStr, ":", 2)
 	if len(originatorFields) != 2 {
-		originatorFields = strings.SplitN(originatorStr, ":", 2)
-		if len(originatorFields) != 2 {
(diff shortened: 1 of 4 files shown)


spiffcs:
cc @kzantow @lumjjb for review =)

spdx/v2/common/package.go
```diff
@@ -73,12 +73,14 @@ func (o *Originator) UnmarshalJSON(data []byte) error {
 	originatorFields := strings.SplitN(originatorStr, ": ", 2)
 
 	if len(originatorFields) != 2 {
-		return fmt.Errorf("failed to parse Originator '%s'", originatorStr)
+		originatorFields = strings.SplitN(originatorStr, ":", 2)
+		if len(originatorFields) != 2 {
+			return fmt.Errorf("failed to parse Originator '%s'", originatorStr)
+		}
```
kzantow, line 79: could we change this to just instead of splitting on `": "`, just modify that to `":"` and `strings.TrimLeft(originatorFields[1], " \t")` below. this way we don't need to have a second split case.

kzantow (approved):

PR #223, opened 2023-09-21
fix(2.2 json): remove emtpy packageVerificationCode

kzantow:
In SPDX 2.2, the `packageVerificationCode` element is optional, but the Go struct does not allow this to be omitted. This was corrected by changing the type to a pointer in SPDX 2.3, but still is an issue in SPDX 2.2. In the case `filesAnalyzed` is `false`, it is _required_ to omit this field. This PR implements a simple method to remove emtpy `packageVerificationCode` entries in SPDX 2.2, specifically.

### DIFF
--- a/spdx/v2/v2_2/json/empty_values_test.go
+++ b/spdx/v2/v2_2/json/empty_values_test.go
@@ -1,0 +1,57 @@
+package json
+
+import (
+	"encoding/json"
+	"github.com/spdx/tools-golang/spdx/v2/common"
+	spdx "github.com/spdx/tools-golang/spdx/v2/v2_2"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func Test_omitsAppropriateProperties(t *testing.T) {
+	tests := []struct {
+		name     string
+		pkg      spdx.Package
+		validate func(t *testing.T, got string)
+	}{
+		{
+			name: "include packageVerificationCode exclusions",
+			pkg: spdx.Package{
+				PackageVerificationCode: common.PackageVerificationCode{
+					ExcludedFiles: []string{},
+				},
+			},
+			validate: func(t *testing.T, got string) {
+				require.Contains(t, got, "packageVerificationCode")
+			},
+		},
+		{
+			name: "include packageVerificationCode value",
+			pkg: spdx.Package{
+				PackageVerificationCode: common.PackageVerificationCode{
+					Value: "1234",
+				},
+			},
+			validate: func(t *testing.T, got string) {
+				require.Contains(t, got, "packageVerificationCode")
+			},
+		},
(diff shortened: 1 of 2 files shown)


spdx/v2/v2_2/package.go
```diff
@@ -122,6 +123,20 @@ type Package struct {
...
+func (p *Package) MarshalJSON() ([]byte, error) {
+	type pkg Package
+	p2 := pkg(*p)
+
+	data, err := json.Marshal(p2)
+	if err != nil {
+		return nil, err
+	}
+
+	return emptyVerificationCodeRegex.ReplaceAll(data, []byte("")), nil
```
lumjjb, line 137: hmm, according to the 2.2 spec, even if this was not empty, it has to be omitted no matter what the value is. In this case, it seems like we would have to check the value and provide an error if it is non-empty as long as `filesAnalyzed=false`?

(on the side, i have a preference from unmarshalling into a map and then into the struct)

kzantow: I updated this with a better decode/delete property/re-encode to remove the property (instead of using a regex, as it was before). I don't really want to alter the behavior much more than this, however. We have a lot of instances that we output whatever data the struct has rather than attempting to always output something according to the spec, so would probably rather not start here but would definitely be open to this in the future, especially once we work on SPDX 3 support.

spdx/v2/v2_2/package.go
```diff
@@ -123,6 +123,30 @@ type Package struct {
...
+func (p Package) MarshalJSON() ([]byte, error) {
+	type pkg Package
+	p2 := pkg(p)
+
+	data, err := marshal.JSON(p2)
+	if err != nil {
+		return nil, err
+	}
+
+	// remove empty packageVerificationCode entries -- required by 2.2 JSON schema but
```
willmurphyscode, line 135: Is there a specific link to the spec you could put here?

spdx/v2/v2_2/json/empty_values_test.go
```diff
@@ -0,0 +1,57 @@
...
+				require.NotContains(t, got, "packageVerificationCode")
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			got, err := json.Marshal(test.pkg)
+			require.NoError(t, err)
+			test.validate(t, string(got))
```
willmurphyscode, line 54: Rather than calling `strings.Contains` on JSON, which feels a bit brittle, would it make sense to unmarshal to a `map[string]any` and assert on the presence/absence and value of specific keys?

willmurphyscode (approved):
LGTM. Thanks for finding a specific link to the spec @kzantow

PR #224, opened 2023-09-21
Stop escaping HTML

kzantow:
This PR modifies the JSON output behavior to prevent HTML escaping within custom `MarshalJSON` calls. By calling `json.Marshal`, the behavior is to use the default of _escaping HTML_ characters. This PR introduces a `marshal.JSON` function that can be used as a drop-in replacement for `json.Marshal` with HTML escaping disabled.

This prevents things like `Author: Keith <keith@example.com>` from getting escaped unnecessarily to something like: `Author: Keith &lt;keith@example.com&gt;`

Depends on #223

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 .PHONY: test
 test: unit fuzz
 
+.PHONY: format
+format:
+	gofmt -w .
+	gosimports -w -local github.com/spdx .
+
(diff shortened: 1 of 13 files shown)


lumjjb (dismissed):
if only we could just set a global flag somewhere, but yea this seems like the best approach for what we have

PR #229, opened 2023-11-03
feat: add external reference category "OTHER"

mcombuechen:
I was missing the External Reference Category "Other" from [Annex F of the 2.3 spec (F.5)](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/). This adds it.


kzantow (approved):
LGTM thanks for the improvement @mcombuechen

PR #240, opened 2024-05-14
DISCUSSION: SPDX 3 data model

kzantow:
DO NOT MERGE!

This PR is intended to provide a spot for an open discussion about what people would like to see for working with SPDX 3 documents. There is a particularly challenging bit, that SPDX 3 is implemented using a single inheritance model, which Go does not directly support. Due to this, I'd like to offer up a few possibilities to determine which users would most like to see.

In the `spdx/v3/v3_0` directory, there are several subdirectories, each implementing a different variant of a model that I believe would support SPDX 3. I've implemented a very simple example with each: creating a couple packages, a file, and some relationships. 👀 look at the `*_test.go` to see the example usage. There, of course, could be certain parts of these different variants used together such as setter errors and first-class embedded structs.

Also check [the README](https://github.com/spdx/tools-golang/pull/240/files#diff-0ba768935a1cf95012e176ff3adb39fc1bc5c9bfb105852bf46a9fa2e97da97e) with some more information and what I see as pros/cons for each of the approaches.

And, of course, if someone has different ideas, please do share them! Perhaps I've missed some obvious solutions that could simplify things here!

There's a high likelihood that I've made some mistakes with the modeling and usage, but this is more to determine the _patterns_ we would like to use, so we can use [the SPDX 3 SHACL model](https://spdx.github.io/spdx-spec/v3.0/model/spdx-model.ttl) to [generate](https://github.com/JPEWdev/shacl2code/pull/9) the corresponding Go code here.

### DIFF
--- a/spdx/v3/v3_0/README.md
+++ b/spdx/v3/v3_0/README.md
@@ -1,0 +1,53 @@
+# SPDX 3 Model Patterns
+
+This is a non-exhaustive list of some patterns we are able to use in Golang to work with an SPDX 3 data model.
+
+I've included some variations I think we should consider. An explanation is as follows, along with some pros/cons:
+
+- [embedded first](embedded_first) - lean into modifying go structs directly with exported fields, using struct embedding.
+This requires using interfaces, but the interfaces are simple single-functions to get an editable embedded struct.
+  - pro: interfaces are simple
+  - pro: modifying structs is simple and idiomatic
+  - pro: minimal surface area: very few additional functions necessary
+  - con: construction is confusing -- which nesting has field X?
+  - con: naming is hard, need `Package` and `IPackage`
+
+
+- [interfaces only](interfaces_only) - only export interfaces and use interfaces to interact with structs.
+NOTE: this also includes a variant that setters return `error`, which could be applied to other options.
+This option only exports the interface types and simple no-arg constructor functions.
+  - pro: idiomatic getter naming (without `Get`)
+  - pro: single type exported, data structs are not
+  - pro: minimal surface area: no extra helper functions/structs that we may decide to change later
+  - con: absolutely no JSON/etc. output
+  - con: getter/setter still not especially idiomatic
+  - con: duplicated code because no embedding
+  - con: construction is tedious and could be error-prone if referencing wrong variable name
+
+
+- [interfaces only with constructors](interfaces_only/constructors)
+  - (generally same pros/cons as above), but:
+  - pro: much simpler construction pattern for users
+  - pro: single validation spot during construction (with returning `error` pattern)
+  - con: more structs and functions exported which cannot be changed without breaking backwards compatibility
+
+
+- [inverted](inverted) - since we know the entire data model and it does not need to be extended,
+  it is possible to implement an uber-struct without the need to use interfaces at all.
+I thought I'd mention this for some semblance of completeness, but this is probably a very bad idea. 
+  - pro: single set of types exported
+  - pro: fairly easy to interact with existing documents
+  - con: easy to construct incorrectly
+  - con: potential unneecssary memory usage
+
+
+- [embedded with interfaces and constructors](embedded_with_interfaces_and_constructors) - takes the approach of both 
+embedding and interfaces, along with constructor functions. This makes creation reasonably simple and working with
+existing documents reasonably simple.
+  - pro: probably the most idiomatic Go of these options
+  - pro: simple construction
+  - pro: reasonably simple to work with existing documents
+  - pro: eliminates some code duplication seen with `interfaces_only`
+  - con: getter/setter still not especially idiomatic
+  - con: larger surface area: at least 4 exported parts for each type
+  - con: absolutely no JSON/etc. output
--- a/spdx/v3/v3_0/embedded_first/model.go
+++ b/spdx/v3/v3_0/embedded_first/model.go
@@ -1,0 +1,161 @@
+package v3_0
+
+import "time"
+
+type ElementID string
+
+// ----------------------- SpdxDocument -----------------------
+
+type SpdxDocument struct {
+	ElementCollection
+
+	NamespaceMap map[string]string
+}
+
+type ISpdxDocument interface {
+	IElementCollection
+}
+
+// ----------------------- ElementCollection -----------------------
+
+type ElementCollection struct {
+	Element
+
+	RootElements []IElement
+	Elements     []IElement
+}
+
+type IElementCollection interface {
+	IElement
+
+	AsElementCollection() *ElementCollection
+}
+
+// ----------------------- Element -----------------------
+
+type Element struct {
+	SpdxID ElementID
+	Name   string
+}
+
+func (e *Element) AsElement() *Element {
+	return e
+}
+
+type IElement interface {
+	AsElement() *Element
+}
+
+var _ interface {
+	IElement
+} = (*Element)(nil)
+
+// ----------------------- Artifact -----------------------
+
+type Artifact struct {
+	Element
+
+	BuiltTime time.Time
+}
+
+func (a *Artifact) AsArtifact() *Artifact {
+	return a
+}
+
+type IArtifact interface {
+	IElement
+
+	AsArtifact() *Artifact
+}
+
+var _ interface {
+	IArtifact
+} = (*Artifact)(nil)
+
+// ----------------------- SoftwareArtifact -----------------------
+
+type SoftwareArtifact struct {
+	Artifact
+
+	CopyrightText string
+}
+
+type ISoftwareArtifact interface {
+	IArtifact
+
+	AsSoftwareArtifact() *SoftwareArtifact
+}
+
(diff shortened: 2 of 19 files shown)


spdx/v3/v3_0/embedded_first/model.go
```diff
@@ -0,0 +1,161 @@
...
+	IElementCollection
+}
+
+// ----------------------- ElementCollection -----------------------
+
+type ElementCollection struct {
+	Element
+
+	RootElements []IElement
+	Elements     []IElement
```
nishakm, line 25: So the problem with this implementation is that RootElement is not class Element but any class that inherits from Element. In Go, I guess, that translates to "any struct that implements the Element interface".

kzantow: Yes, in this variant, the `I*` types are all `interfaces` -- so these are "any types implementing the `IElement` interface". These could be named differently... naming is hard! there doesn't seem to be a way to avoid interfaces unless we made any reference field an `any` type, but then there's no compiler enforcement of elements going in there. (Also I probably modeled this wrong, I was going by the [software profile svg](https://raw.githubusercontent.com/spdx/spdx-3-model/main/model%20Software.svg) which appears to have `*` for both of these fields).

kzantow: FWIW, I _think_ personally I'm leaning towards the `embedded_with_interfaces_and_constructors` variant, with setters that return `error`, [as the example currently shows](https://github.com/spdx/tools-golang/pull/240/files/c8d843ccbcda6cf5128a628e42da75a53601d228#diff-1f2cba19eca292de24e0f81517e713ff958d56252243c2c4192359fdf2aaa0d3R13). This allows generated code to enforce valid values for all fields and for a user to most quickly understand when things failed rather than needing to run a full validation once all values are set. It also eliminates the need to export the data structs (all the embedded variants need exported structs)

nishakm: Oh my bad: the property is of type IElement

nishakm: This seems to work. It would be easier to generate:
```
package main

import "fmt"

type ElementP struct {
	Name string
}

type Element interface {
	AsElement() ElementP
}

type Artifact struct {
	ElementP
	StandardName []string
}

func (a Artifact) AsElement() ElementP {
	return a.ElementP
}

type ElementCollection struct {
	ElementP
	RootElement Element
}

func main() {
	a := Artifact{
		ElementP: ElementP{
			Name: "AnArtifact",
		},
		StandardName: []string{"TheArtifact"},
	}

	ec := ElementCollection{
		ElementP: ElementP{
			Name: "AnElementCollection",
		},
		RootElement: a,
	}

	fmt.Println(ec)
}
```

nishakm: One more issue: The jinja renderer I am using from the shacl2code project can only generate one file. So if we were to generate a model and a constructor file, we would need to roll our own jinja renderer...which I would rather do once I can handle the code better. You are welcome to try building one that will generate multiple files.

kzantow: One note: there is no requirement to have separate files -- I just did this to separate things logically, but all the code for each of these approaches could be generated in a single file. In other words: if we were to export all 4 types: interface, data struct, constructor, constructor props; all of these could be one after the other in the same file.

From the "generic shacl2code" perspective, I think in the example there's a bit of an inconsistency with the example you posted: only a single struct type gets renamed (`ElementP`) for an `Element` interface, but there are no other interfaces created. I think the easiest thing to do in `shacl2code` is probably: generate an exported interface and a struct for each type. It also probably makes the most sense to name the _interface_ after the shacl type rather than the struct, as one of my examples has. Abstract types are a question: golang would let you indiscriminately construct any abstract type if these structs are exported, and they would need to be generated and exported to be used like superclasses through embedding.

The `interfaces_only` approach has the potential to model inheritance defined in shacl in the most correct way, I believe -- because abstract classes can _omit_ constructor functions and since the types themselves are not exported, there's no way for a user to create these types incorrectly. It has some complexity that at the time you're generating code for a type, it needs to have access to supertypes to be able to inline all the properties -- a quick look makes me think this wouldn't be too hard to accomplish, since it looks like we can [get a reference to a parent class](https://github.com/JPEWdev/shacl2code/pull/9/files#diff-3d82489614fc2a7b7f166bce8660b271d2224754c390411135fd294ea15c3d4bR92) and [jinja supports recursion](https://stackoverflow.com/questions/32771924/macro-recursion-in-jinja2).

An aside: _naming is hard_! Is the `P` from `ElementP` short for `Props` or `Properties`? I am sorta partial to `Data` for these structs despite using `Impl` in one of the variants, but also like `Props`, at least for the constructor props types, or other things; happy for any other suggestions but would definitely like to keep it short but maybe more descriptive than a single letter. Interfaces only allows the lowercase name to be used, since it's not exported and we don't have to come up with names, so I guess there's another plus for that one...

spdx/v3/v3_0/interfaces_only/model.go
```diff
@@ -0,0 +1,342 @@
...
+
+import "time"
+
+type ElementID string
+
+type SpdxDocument interface {
+	ElementCollection
+}
+
+type spdxDocument struct {
```
nishakm, line 11: This is the "interfaces only" example, but this is a struct. Is a concrete class a struct and an abstract class an interface?

spdx/v3/v3_0/interfaces_only/model.go
```diff
@@ -0,0 +1,342 @@
...
+	// Element properties
+	spdxID ElementID
+	name   string
```
nishakm, line 14: Generating each abstract parents' properties using jinja would be ok to do. You think this is a better option than using a struct called `elementProperties` (purposefully not exporting the struct, but making the concrete class implement the interface `Element`).

spdx/v3/v3_0/embedded_first/model.go
nishakm: I think I agree with you on going with the interface only approach i.e. Abstract class -> Go Interface with getter and setter for each property. Concrete class -> Go struct which implements the getter and setter methods for the abstract class.

I personally would prefer having corresponding internal structs to hold common properties in the concrete classes, but it's not a deal breaker. We have to embed the interfaces anyway (Artifact is actually an abstract class that inherits from Element).

spdx/v3/v3_0/embedded_with_interfaces_and_constructors/model.go
```diff
@@ -0,0 +1,254 @@
...
+	Name() string
+	SetName(name string)
+}
+
+type ElementImpl struct {
+	spdxID ElementID
+	name   string
+}
+
+var _ Element = (*ElementImpl)(nil)
```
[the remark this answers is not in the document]
kzantow: This is a way to write `implements <interface>` in go:
```
var _ Element = (*ElementImpl)(nil)
```
is a compile-time check that an `*ElementImpl` type can be assigned to an `Element` interface, it gets optimized out by the compiler so no runtime overhead is incurred. It's not really necessary for any of the code generation, I mostly added it for my own benefit.

A similar compile-time check would happen with constructor functions, like:
```
func NewPackage() Package {
   return &packageImpl{}
}
```
will similarly fail to compile if `packageImpl` doesn't adhere to the Package interface

spdx/v3/v3_0/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+- [embedded with interfaces and constructors](embedded_with_interfaces_and_constructors) - takes the approach of both 
+embedding and interfaces, along with constructor functions. This makes creation reasonably simple and working with
+existing documents reasonably simple.
+  - pro: probably the most idiomatic Go of these options
+  - pro: simple construction
+  - pro: reasonably simple to work with existing documents
+  - pro: eliminates some code duplication seen with `interfaces_only`
+  - con: getter/setter still not especially idiomatic
+  - con: larger surface area: at least 4 exported parts for each type
```
lumjjb, line 52: hmm... yea i wonder if there's anyhting we can do with getters/setters here with reflection...

why do we need to export the internal struct again? can you remind me?

lumjjb (commented):
Thank you sooo much @kzantow for this super detailed rundown - I like embedded with interfaces and constructors the best!

goneall:
> @goneall is there a JSON schema spec for v3? Also would love your thoughts on the proposal since you've been doing this with Java as well (since i imagine having inheritance is nice for this data model :))

Yes - https://spdx.github.io/spdx-spec/v3.0/model/schema.json - This is generated from the SHACL file using the code generator Joshua wrote.  I haven't reviewed the above discussion yet, but I thought I'd do a quick reply with a pointer to the schema

nishakm:
Following up: any consensus on what direction we want to take?

kzantow:
Thanks everyone for the feedback here! After some experimentation and back and forth with @nishakm, I have a proposed direction and would love feedback there. I opened a new PR for this, feel free to try it out: https://github.com/spdx/tools-golang/pull/247

Issue #117, opened 2022-03-26
Add tests to CI for Windows (and maybe others)


swinslow:
Since Windows handles filepaths differently (and probably some other things I'm not thinking of), we should also configure the GitHub Actions testing to run on a Windows VM -- if that's doable within GitHub Actions. Should also look into whether to do the same for e.g. MacOS and/or others.

CatalinStratu:
@swinslow , I can try to handle this issue!

swinslow:
That's great -- thank you @CatalinStratu!

CatalinStratu:
Thanks, I'll take care of this task after I'm done with task #120

CatalinStratu:
It will be necessary to refactor some unit tests. To be able to make all the tests run well on both Linux and Windows
![image](https://user-images.githubusercontent.com/28156781/161084556-349afd8a-6816-42c4-880b-2b523ce39de8.png)

swinslow:
Thanks @CatalinStratu. If there are some tests that need refactoring, that makes sense and that's fine to include in a proposed PR.

I'd suggest breaking up refactoring changes into small separate PRs wherever possible so that they can be discussed one by one, rather than one very large PR which would be harder to evaluate.

On the specific failures from your screenshot, note that the verification code and number of Files should be the same regardless of whether the Golang tools are being run on Windows or on Linux. The fact that they're different here is a good thing; that means that the tests are identifying a problem with the tools-golang that is leading to different results on the different platforms. We should dig into why the results are different, and figure out how to fix the bug in the tools -- rather than changing the tests to hide the bug  :)

For this particular test, it looks like this is where the `builder` package is creating an SPDX document by analyzing a test folder. This particular test folder, [`testdata/project1`](https://github.com/spdx/tools-golang/tree/main/testdata/project1), contains a symbolic link file. I'd be willing to bet that's why you're seeing different results on Windows vs. Linux, since (as far as I know) Windows doesn't have a symbolic link concept.

Can you submit a separate issue for this particular test error, so that we can dig into that specifically? Resolving that bug is going to take a bit of digging to see how symbolic links are handled for other SPDX tools, and it's a bit separate from this issue #117 for enabling the Windows tests to run within the GitHub Actions CI system. Thank you!

swinslow:
[...]

PR #241, opened 2024-05-16
fix(windows): crash on out of bounds for root path cleanup

manifestori:
I am using this lib in a tool I've been contributing to. 

```go
import builder_common "github.com/spdx/tools-golang/builder"

cwd, err := os.Getwd()
if err != nil {
	return nil, err
}

...
ref := &builder.Config{
		NamespacePrefix: "https://spdx.org/spdxdocs/",
		CreatorType:     "Tool",
		Creator:         "me",
}


builder_common.Build("mypackage", cwd, ref)
```

On Linux - no issues. 
On Windows - I get out of bounds. Also, I tried running tests (without any modifications), but they fail (due to Windows path issues). 
I tried my best to test it in the given state.

Also, note that I changed `"."` to `".\\"` .
if a root path has "/" for suffix (ie. my\path\ ) this will not impose a problem ( .\\my\file.txt equals '.\my\file.txt ). having only "." is problematic for cases where the root path includes `\` and the rel path is just .file.txt and not .\file.txt

Im not a Windows Pro but this seems right

### DIFF
--- a/builder/build_package.go
+++ b/builder/build_package.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"strings"
 
 	"github.com/spdx/tools-golang/spdx"
 	"github.com/spdx/tools-golang/spdx/v2/common"
@@ -42,7 +43,11 @@ func BuildPackageSection(packageName string, dirRoot string, pathsIgnore []strin
 	for _, fp := range filepaths {
 		newFilePatch := ""
 		if osType == "windows" {
-			newFilePatch = filepath.FromSlash("." + fp[dirRootLen:])
+			if strings.HasPrefix(fp, dirRoot) {
+				newFilePatch = filepath.FromSlash("." + fp[dirRootLen:])
+			} else {
+				newFilePatch = filepath.FromSlash("." + fp)
+			}
 		} else {
 			newFilePatch = filepath.FromSlash("./" + fp)
 		}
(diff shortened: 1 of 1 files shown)


kzantow:
Hi @manifestori; thanks for the contribution! It looks to me like this library should support Windows for some of this functionality, so I think we should make sure these code paths work by running the tests on Windows, too. I thought I'd give that a try, so I made [a PR with a simple workflow to also run the existing tests on Windows](https://github.com/spdx/tools-golang/pull/242). Unfortunately, it looks like there are some [definite](https://github.com/spdx/tools-golang/actions/runs/9118676741/job/25072334206?pr=242#step:4:111) [incompatibilities](https://github.com/spdx/tools-golang/actions/runs/9118676741/job/25072334206?pr=242#step:4:1724) at least between the _tests_ and Windows. I looked in to this last night a bit, and as expected, this is primarily due to path handling, but the fix in _this PR_ also had some unexpected behavior where tests, at least, got double slashes e.g. `.\\something` rather than `.\something`. I don't use this functionality myself, but I'd like to try to make sure it works in expected ways for everyone. I'd love feedback about how to best handle Windows moving forward. I posed a question about the main issues I found when investigating here, if you had any thoughts, please do add them: https://github.com/spdx/tools-golang/issues/117#issuecomment-2117799417

manifestori:
Yeah, as I expected `.\\someting` vs `.\something` I can make sure to add the correct prefix (`.` or `.\`) 
they are equivalent, but I see why we would want to differentiate them. 

Currently, I had to issue a prompt fix as this is currently broken; I will be using my fork until we resolve this, but I suggest we issue this as 0.5.5 

I'll be happy to contribute to the test discussions as well, especially with the up-and-coming spdx3.0 support.

manifestori:
@kzantow added trim to double slashes if input ended up having double slash (`".\\\\"` with escapes). 
They are semantically equivalent but syntactically different anyway. wdyt?

manifestori:
closing in favor of #242

PR #242, opened 2024-05-16
chore: add windows test runner

kzantow:
Since there is Windows-specific code, the tests should also run on Windows.

Fixes: #117

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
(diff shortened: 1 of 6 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
kzantow, line 1: This seems logical, since `go fmt` produces only LF endings and if the `.go` files get checked out on windows with CRLF, there are "every file changed" when running `go fmt`.

manifestori (approved):
much better then my quick workaround, thank you for resolving this. 

LGTM

kzantow:
Thanks @manifestori! I'll try to get a release done this week, but in the meantime are you able to use `main`?

PR #247, opened 2024-08-01
feat: prototype v3_0 model

kzantow:
This PR implements one possible data model to support SPDX 3, along with functional JSON serialization / deserialization. There are sure to be some rough edges!

The easiest way to understand what usage looks like is to check out the tests -- they perform some basic document creation, along with serializing, deserializing, and re-serializing to verify the documents are the same.

A few rough edges with SPDX 3 in general seem to be:
* all elements are required to have a _creation Info_; this PR includes a feature during JSON serialization to set all elements creation info if it is unset, I don't know if this is a great idea but it would make things a lot easier to deal with...
* it could be possible to have multiple SPDX document objects in the same json document
* having an SPDX document and an SBOM independently is somewhat confusing
* there's a chicken-and-egg problem between _creation info_ and _agent_, where both are required and both reference each other, so some helper as I've implemented here would probably be good; maybe it should take different information

This PR does _not_ add anything in the way of data validation, so it is probably pretty easy to create invalid documents. How to handle this is TBD.

This PR also does _not_ add any conversion functionality to/from older versions of SPDX.

Feedback strongly encouraged!


pandatix (changes requested):

spdx/v3/v3_0/spdx_test.go
```diff
@@ -0,0 +1,214 @@
...
+			Name:         "not-tools-golang",
+		}})
+
+	sbom := &spdx.SoftwareSbom{}
+	doc.RootElements.Append(sbom)
+
+	// create a package
+
+	pkg1 := &spdx.SoftwarePackage{
+		SoftwareSoftwareArtifact: spdx.SoftwareSoftwareArtifact{Artifact: spdx.Artifact{Element: spdx.Element{
```
kzantow, line 35: I'm just going to call out the two least desirable traits of this proposal: moderately deep inheritance heirarchies with embedding are not especially fun to work with in initializers, since you must nest embedded types like this to set the "parent" type properties.

spdx/v3/v3_0/spdx_test.go
```diff
@@ -0,0 +1,214 @@
...
+			Name: "some-package-1",
+		}}},
+		SoftwarePackageVersion: "1.2.3",
+	}
+
+	// create another package
+
+	pkg2 := &spdx.AiAIPackage{}
+	pkg2.Name = "some-package-2"
+	pkg2.SoftwarePackageVersion = "2.4.5"
```
kzantow, line 45: But it's also possible to avoid the nesting by accessing properties like this.

spdx/v3/v3_0/spdx_test.go
```diff
@@ -0,0 +1,214 @@
...
+	doc.RootElements.Append(sbom)
+
+	// create a package
+
+	pkg1 := &spdx.SoftwarePackage{
+		SoftwareSoftwareArtifact: spdx.SoftwareSoftwareArtifact{Artifact: spdx.Artifact{Element: spdx.Element{
+			Name: "some-package-1",
+		}}},
+		SoftwarePackageVersion: "1.2.3",
+	}
```
kzantow, line 39: I'm just going to call out the two least desirable traits of this proposal: moderately deep inheritance heirarchies with embedding are not especially fun to work with in initializers, since you must nest embedded types like this to set the "parent" type properties.

spdx/v3/v3_0/spdx_test.go
```diff
@@ -0,0 +1,214 @@
...
+	//	t.Errorf(text)
+	//}
+
+	// some basic usage:
+
+	var pkgs []*spdx.SoftwarePackage
+	for _, sbom := range doc.RootElements.SoftwareSbomIter() {
+		for _, rel := range sbom.RootElements.RelationshipIter() {
+			if rel.RelationshipType == spdx.RelationshipType_Contains {
+				spdx.As(rel.From, func(f *spdx.SoftwareFile) {
```
kzantow, line 126: The second thing is accessing single-instance properties like this. There is no magic typing that could be done to add methods like `rel.From.SoftwareFile()` to get a SoftwareFile, for example, like is done with the specialized collections. This means it's required to type assert... but because of the inheritance, you cannot type assert to `*spdx.SoftwareFile`, or you'll miss any subtypes, so the `spdx.As` function is a helper that simply switches on the type provided -- in this case `*spdx.SoftwareFile` switches to test for `spdx.AnySoftwareFile` interface, and if found calls `.asSoftwareFile()` to get a reference. This allows the inheritance to work properly at the expense of being able to do anything idiomatic here.

spdx/v3/v3_0/spdx_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// some basic usage:
+
+	var pkgs []*spdx.SoftwarePackage
+	for _, sbom := range doc.RootElements.SoftwareSbomIter() {
+		for _, rel := range sbom.RootElements.RelationshipIter() {
+			if rel.RelationshipType != spdx.RelationshipType_Contains {
+				continue
+			}
+			_ = spdx.As(rel.From, func(f *spdx.SoftwareFile) any {
```
kzantow, line 85: The second thing is accessing single-instance properties like this. There is no magic typing as is done with the specialized collections that could be done to add methods like rel.From.SoftwareFile(), for example. This means it's required to type assert... but because of the inheritance, you cannot type assert to *spdx.SoftwareFile, or you'll miss any subtypes, so the spdx.As function is a helper that simply switches on the type provided -- in this case *spdx.SoftwareFile switches to test for spdx.AnySoftwareFile interface, and if found calls .asSoftwareFile() to get a reference. This allows the inheritance to work properly at the expense of not really being able to do anything idiomatic here.

spdx/v3/v3_0_1/spdx_serialization_test.go
```diff
@@ -32,8 +32,8 @@ func Test_spdxExportImportExport(t *testing.T) {
 			&Tool{Element: Element{
 				ExternalIdentifiers: ExternalIdentifierList{
 					&ExternalIdentifier{
-						ExternalIdentifierType: ExternalIdentifierType_Cpe23,
-						Identifier:             "cpe23:a:myvendor:my-product:*:*:*:*:*:*:*",
+						Type:       ExternalIdentifierType_Cpe23,
+						Identifier: "cpe23:a:myvendor:my-product:*:*:*:*:*:*:*",
```
pandatix: Shouldn't it be `cpe:2.3:a:myvendor:my-product:*:*:*:*:*:*:*:*` (see prefix) ?

PR #249, opened 2024-11-14
DO NOT MERGE: SPDX 3 Bindings

JPEWdev:
This is a *very* initial pass at the generated go bindings from shacl2code. These bindings are incomplete, but give a basic idea of what they might look like for evaluation purposes

### DIFF
--- a/spdx/v3/v3_0_1/model.go
+++ b/spdx/v3/v3_0_1/model.go
@@ -1,0 +1,16878 @@
+//
+//
+//
+
+package v3_0_1
+
+import (
+    "encoding/json"
+    "fmt"
+    "reflect"
+    "regexp"
+    "sort"
+    "strconv"
+    "strings"
+    "time"
+
+    "github.com/ncruces/go-strftime"
+)
+
+// Validation Error
+type ValidationError struct {
+    Property string
+    Err string
+}
+
+func (e *ValidationError) Error() string { return e.Property + ": " + e.Err }
+
+// Conversion Error
+type ConversionError struct {
+    From string
+    To string
+}
+
+func (e *ConversionError) Error() string {
+    return "Unable to convert from " + e.From + " to " + e.To
+}
+
+// Decode Error
+type DecodeError struct {
+    Path Path
+    Err string
+}
+
+func (e *DecodeError) Error() string {
+    return e.Path.ToString() + ": " + e.Err
+}
+
+type EncodeError struct {
+    Path Path
+    Err string
+}
+
+func (e *EncodeError) Error() string {
+    return e.Path.ToString() + ": " + e.Err
+}
+
+// Path
+type Path struct {
+    Path []string
+}
+
+func (p *Path) PushPath(s string) Path {
+    new_p := *p
+    new_p.Path = append(new_p.Path, s)
+    return new_p
+}
+
+func (p *Path) PushIndex(idx int) Path {
+    return p.PushPath("[" + strconv.Itoa(idx) + "]")
+}
+
+func (p *Path) ToString() string {
+    return "." + strings.Join(p.Path, ".")
+}
+
+// Error Handler
+type ErrorHandler interface {
+    HandleError(error, Path)
+}
+
+// Reference
+type Ref[T SHACLObject] interface {
+    GetIRI() string
+    GetObj() T
+    IsSet() bool
+    IsObj() bool
+    IsIRI() bool
+}
+
+type ref[T SHACLObject] struct {
+    obj *T
+    iri string
+}
+
+func (r ref[T]) GetIRI() string {
+    if r.iri != "" {
+        return r.iri
+    }
+    if r.obj != nil {
+        o := *r.obj
+        if o.ID().IsSet() {
+            return o.ID().Get()
+        }
+    }
+    return ""
+}
+
+func (r ref[T]) GetObj() T {
+    return *r.obj
+}
+
+func (r ref[T]) IsSet() bool { return r.IsIRI() || r.IsObj() }
+func (r ref[T]) IsObj() bool { return r.obj != nil }
+func (r ref[T]) IsIRI() bool { return r.iri != "" }
+
+func MakeObjectRef[T SHACLObject](obj T) Ref[T] {
+    return ref[T]{&obj, ""}
+}
+
+func MakeIRIRef[T SHACLObject](iri string) Ref[T] {
+    return ref[T]{nil, iri}
+}
+
+// Convert one reference to another. Note that the output type is first so it
+// can be specified, while the input type is generally inferred from the argument
+func ConvertRef[TO SHACLObject, FROM SHACLObject](in Ref[FROM]) (Ref[TO], error) {
+    if in.IsObj() {
+        out_obj, ok := any(in.GetObj()).(TO)
+        if !ok {
+            return nil, &ConversionError{reflect.TypeOf(ref[FROM]{}).Name(), reflect.TypeOf(ref[TO]{}).Name()}
+        }
+        return ref[TO]{&out_obj, in.GetIRI()}, nil
+    }
+    return ref[TO]{nil, in.GetIRI()}, nil
+}
+
+type Visit func(Path, any)
+
+// Base SHACL Object
+type SHACLObjectBase struct {
+    // Object ID
+    id Property[string]
+    typ SHACLType
+    typeIRI string
+}
+
+func (self *SHACLObjectBase) ID() PropertyInterface[string] { return &self.id }
+
+func (self *SHACLObjectBase) Validate(path Path, handler ErrorHandler) bool {
+    var valid bool = true
+
+    switch self.typ.GetNodeKind() {
+    case NodeKindBlankNode:
+        if self.ID().IsSet() && ! IsBlankNode(self.ID().Get()) {
+            handler.HandleError(&ValidationError{
+                "id",
+                "ID must by be blank node"},
+                path.PushPath("id"))
+            valid = false
+        }
+    case NodeKindIRI:
+        if ! self.ID().IsSet() || ! IsIRI(self.ID().Get()) {
+            handler.HandleError(&ValidationError{
+                "id",
+                "ID must be an IRI"},
+                path.PushPath("id"))
+            valid = false
+        }
+    case NodeKindBlankNodeOrIRI:
+        if self.ID().IsSet() && ! IsBlankNode(self.ID().Get()) && ! IsIRI(self.ID().Get()) {
+            handler.HandleError(&ValidationError{
+                "id",
+                "ID must be a blank node or IRI"},
+                path.PushPath("id"))
+            valid = false
+        }
+    default:
+        panic("Unknown node kind")
+    }
+
+    return valid
+}
+
+func (self *SHACLObjectBase) Walk(path Path, visit Visit) {
+    self.id.Walk(path, visit)
+}
+
+func (self *SHACLObjectBase) EncodeProperties(data map[string]interface{}, path Path, state *EncodeState) error {
+    if self.typeIRI != "" {
+        data["type"] = self.typeIRI
+    } else {
+        data["type"] = self.typ.GetCompactTypeIRI().GetDefault(self.typ.GetTypeIRI())
+    }
+
+    id_prop := self.typ.GetIDAlias().GetDefault("@id")
+
+    if self.id.IsSet() {
+        val, err := EncodeIRI(self.id.Get(), path.PushPath(id_prop), map[string]string{}, state)
+        if err != nil {
+            return err
+        }
+        data[id_prop] = val
+    }
+
+    return nil
+}
+
+func (self *SHACLObjectBase) GetType() SHACLType {
+    return self.typ
+}
+
+func (self *SHACLObjectBase) setTypeIRI(iri string) {
+    self.typeIRI = iri
+}
+
+func ConstructSHACLObjectBase(o *SHACLObjectBase, typ SHACLType) *SHACLObjectBase {
+    o.id = NewProperty[string]("id", []Validator[string]{ IDValidator{}, })
+    o.typ = typ
+    return o
+}
+
+type SHACLObject interface {
+    ID() PropertyInterface[string]
+    Validate(path Path, handler ErrorHandler) bool
+    Walk(path Path, visit Visit)
+    EncodeProperties(data map[string]interface{}, path Path, state *EncodeState) error
+    GetType() SHACLType
+    setTypeIRI(iri string)
+}
+
+func EncodeSHACLObject(o SHACLObject, path Path, state *EncodeState) (any, error) {
+    if state != nil {
+        if state.Written[o] {
+            if o.ID().IsSet() {
+                return o.ID().Get(), nil
+            }
+
+            return nil, &EncodeError{
+                path,
+                "Object referenced multiple times, but does not have an ID assigned",
+            }
+        }
+
+        state.Written[o] = true
+    }
+
+    d := make(map[string]interface{})
+    return d, o.EncodeProperties(d, path, state)
+}
+
+// Extensible Object
+
+type SHACLExtensibleBase struct {
+    properties map[string][]any
+}
+
+func (self *SHACLExtensibleBase) GetExtProperty(name string) []any {
+    return self.properties[name]
+}
+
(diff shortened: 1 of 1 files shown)


nishakm:
I'd like to try and pair this down to the basics required to create an SPDX "document". So far, I've been operation on the assumption that this requires:
- Agent
- CreationInfo

For example, in example1 of the [spdx-examples](https://github.com/spdx/spdx-examples/blob/master/software/example1/spdx3.0/example1.json) repo, the bare minimum required json is:
```
{
  "@context" : "https://spdx.org/rdf/3.0.1/spdx-context.jsonld",
  "@graph" : [ {
    "@id" : "_:creationInfo_0",
    "type" : "CreationInfo",
    "specVersion" : "3.0.1",
    "createdBy" : [ "https://swinslow.net/spdx-examples/example1/hello-v3-specv3/SPDXRef-gnrtd0" ],
    "created" : "2021-08-26T01:46:00Z"
  }, {
    "spdxId" : "https://swinslow.net/spdx-examples/example1/hello-v3-specv3/SPDXRef-gnrtd0",
    "type" : "Person",
    "externalIdentifier" : [ {
      "type" : "ExternalIdentifier",
      "identifier" : "steve@swinslow.net",
      "externalIdentifierType" : "email"
    } ],
    "name" : "Steve Winslow",
    "creationInfo" : "_:creationInfo_0"
  }, {
    "spdxId" : "https://swinslow.net/spdx-examples/example1/hello-v3-specv3/document0",
    "type" : "SpdxDocument",
    "rootElement" : [ "https://swinslow.net/spdx-examples/example1/hello-v3-specv3/SPDXRef-gnrtd0" ],
    "name" : "hello",
    "creationInfo" : "_:creationInfo_0"
  }]
}
```
I'm not sure if creating an SPDX document requires a package, but from my experimentation, it does require a CreationInfo and an Agent to be created beforehand. In order to create the CreationInfo, the Agent's SPDXID needs to be created first.
Does this match everyone's understanding?

JPEWdev:
Yes, that looks correct to me.

goneall:
Me too - that's about the minimum

nishakm:
First pass feedback:
- type ([T any]) parameter is only allowed for go 1.18 and higher and the current go.mod says go 1.13. I would drop type parameters.
- I'm still trying to figure out how to actually generate the json I posted above. How do I create a "Person" with a name and email?
- I noticed some weird combination of type constraints and interfaces. The purpose of using the interface is because Go doesn't understand polymorphism. In order to make a "is a" relationship, you have to make an interface and then have objects implement that interface.

nishakm:
Some initial testing:
```
  1 package main
  2 
  3 import "spdx3/pkg/spdx"
  4 
  5 func main() {
  6         p := spdx.MakePerson()
  7         p.Name = "Steve Winslow"
  8 }
```
->
 go build main.go 
# command-line-arguments
./main.go:7:2: cannot assign to p.Name (neither addressable nor a map index expression)

I'm actually surprised the compiler can find that the `Person` struct has a property called `Name` because I couldn't find it in my code search.
```
./main.go:7:4: p.name undefined (type spdx.Person has no field or method name, but does have Name)
```

JPEWdev:
> First pass feedback:
> 
>     * type ([T any]) parameter is only allowed for go 1.18 and higher and the current go.mod says go 1.13. I would drop type parameters.

I'll see if I can make something work. Is there a specific minimum version of go we should support? 1.13 feels very old as a minimum version. If there is a way to allow the "generic" types in 1.13 I'm fine with that but the bindings will be.... pretty cantankerous if there isn't a way to do this.

> 
>     * I'm still trying to figure out how to actually generate the json I posted above. How do I create a "Person" with a name and email?

Something like:
```go
p := model.MakePerson()
p.Name().Set("Joshua")
email = model.MakeExternalIndentifier()
email.ExternalIdentifierType().Set(model.ExternalIdentifierTypeEmail)
email.Identifier().Set("noreply@noreply.com")

# I might add an append to add a single item to the end, because that is common
p.ExternalIdentifier().Set([]Ref[ExternalIdentifier]{model.MakeObjectRef(email)})
```
> 
>     * I noticed some weird combination of type constraints and interfaces. The purpose of using the interface is because Go doesn't understand polymorphism. In order to make a "is a" relationship, you have to make an interface and then have objects implement that interface.

Can you be more specific? In general, you should only be dealing with interfaces. 

> 
>     * No Getters and Setters for properties in objects.

That's really not possible if we want to do any sort of assignment validation. The assignment validation is _really_ helpful in the other language bindings since it gives immediate feedback about where you're using the model wrong, so I think the bindings are much more useful with them, even if it's not "the go way". We really need to be dealing with all the SPDX objects via interfaces to keep the polymorphism as best possible (at least as far as I can tell), and you can't (AFAICT) have properties on interfaces, just getters and setters.

nishakm:
> > First pass feedback:
> > ```
> > * type ([T any]) parameter is only allowed for go 1.18 and higher and the current go.mod says go 1.13. I would drop type parameters.
> > ```
> 
> I'll see if I can make something work. Is there a specific minimum version of go we should support? 1.13 feels very old as a minimum version. If there is a way to allow the "generic" types in 1.13 I'm fine with that but the bindings will be.... pretty cantankerous if there isn't a way to do this.
> 

The way to do this is just not use `type PropertyInterface[T any] interface` as [T any] translates to "interface" anyway. `type PropertyInterface interface` is enough.


> 
> Can you be more specific? In general, you should only be dealing with interfaces.

Oh I see my problem: in some places interfaces are called "Element", "Agent" etc, and in others, it's called "PropertyInterface" rather than "Property". But now I know where to look for things. Maybe consistent naming for the SHACL parts and the SPDX parts?

> 
> > ```
> > * No Getters and Setters for properties in objects.
> > ```
> 
> That's really not possible if we want to do any sort of assignment validation. The assignment validation is _really_ helpful in the other language bindings since it gives immediate feedback about where you're using the model wrong, so I think the bindings are much more useful with them, even if it's not "the go way". We really need to be dealing with all the SPDX objects via interfaces to keep the polymorphism as best possible (at least as far as I can tell), and you can't (AFAICT) have properties on interfaces, just getters and setters.

I got it - manipulating objects happen at the SHACL level so that validation works. Let me try again.

JPEWdev:
> > > First pass feedback:
> > > ```
> > > * type ([T any]) parameter is only allowed for go 1.18 and higher and the current go.mod says go 1.13. I would drop type parameters.
> > > ```
> > 
> > 
> > I'll see if I can make something work. Is there a specific minimum version of go we should support? 1.13 feels very old as a minimum version. If there is a way to allow the "generic" types in 1.13 I'm fine with that but the bindings will be.... pretty cantankerous if there isn't a way to do this.
> 
> The way to do this is just not use `type PropertyInterface[T any] interface` as [T any] translates to "interface" anyway. `type PropertyInterface interface` is enough.

Hmm, I don't think that's the same. I'm using Generics here: https://go.dev/doc/tutorial/generics which it looks like were added in go 1.18.... it would be pretty annoying to write this code without them TBH (it would have to be way more verbose)

> 
> > Can you be more specific? In general, you should only be dealing with interfaces.
> 
> Oh I see my problem: in some places interfaces are called "Element", "Agent" etc, and in others, it's called "PropertyInterface" rather than "Property". But now I know where to look for things. Maybe consistent naming for the SHACL parts and the SPDX parts?
> 
> > > ```
> > > * No Getters and Setters for properties in objects.
> > > ```
> > 
> > 
> > That's really not possible if we want to do any sort of assignment validation. The assignment validation is _really_ helpful in the other language bindings since it gives immediate feedback about where you're using the model wrong, so I think the bindings are much more useful with them, even if it's not "the go way". We really need to be dealing with all the SPDX objects via interfaces to keep the polymorphism as best possible (at least as far as I can tell), and you can't (AFAICT) have properties on interfaces, just getters and setters.
> 
> I got it - manipulating objects happen at the SHACL level so that validation works. Let me try again.

nishakm:
> Something like:
> 
> ```go
> p := model.MakePerson()
> p.Name().Set("Joshua")
> email := model.MakeExternalIndentifier()
> email.ExternalIdentifierType().Set(model.ExternalIdentifierTypeEmail)
> email.Identifier().Set("noreply@noreply.com")
> 
> # I might add an append to add a single item to the end, because that is common
> p.ExternalIdentifier().Set([]Ref[ExternalIdentifier]{model.MakeObjectRef(email)})
> ```

Can you finish this with CreationInfo? I am having trouble with the `ListPropertyInterface`.

JPEWdev:
Oh, right, the typing is a little weird with the references in golang.... hang on and I'll get an example that works.

JPEWdev:
@nishakm Hey, you found a bug :) ! The latest version works with this example code:

```go
package main

import (
    "time"
    "os"
    "fmt"
    "encoding/json"
    "model"
)

func main() {
    // Note: Object de-duplication when serializing is not implemented in the
    // go bindings yet, so this example code links by IRI reference instead of
    // by object reference. This would not normally be the recommended way to
    // do this, as object references are much more usable

    p := v3_0_1.MakePerson()
    p.ID().Set("http://spdx.org/spdxdocs/person-example")
    p.Name().Set("Joshua")

    ci := v3_0_1.MakeCreationInfo()

    // This would not normally be required, but references are done by IRI
    // because deduplication is not implemented
    ci.ID().Set("_:creation-info")

    ci.SpecVersion().Set("3.0.1")

    // Normally would do:
    // ci.CreatedBy().Set([]v3_0_1.Ref[v3_0_1.Agent]{v3_0_1.MakeObjectRef[v3_0_1.Agent](p)})
    ci.CreatedBy().Set([]v3_0_1.Ref[v3_0_1.Agent]{v3_0_1.MakeIRIRef[v3_0_1.Agent](p.ID().Get())})
    ci.Created().Set(time.Now())

    // Normally would do:
    //p.CreationInfo().Set(v3_0_1.MakeObjectRef(ci))
    p.CreationInfo().Set(v3_0_1.MakeIRIRef[v3_0_1.CreationInfo](ci.ID().Get()))

    email := v3_0_1.MakeExternalIdentifier()
    email.ExternalIdentifierType().Set(v3_0_1.ExternalIdentifierTypeEmail)
    email.Identifier().Set("noreply@noreply.com")

    p.ExternalIdentifier().Set([]v3_0_1.Ref[v3_0_1.ExternalIdentifier]{v3_0_1.MakeObjectRef(email)})

    doc := v3_0_1.MakeSpdxDocument()
    doc.ID().Set("http://spdx.org/spdxdocs/doc-example")

    // Normally would do:
    //doc.CreationInfo().Set(v3_0_1.MakeObjectRef(ci))
    doc.CreationInfo().Set(v3_0_1.MakeIRIRef[v3_0_1.CreationInfo](ci.ID().Get()))

    objset := v3_0_1.NewSHACLObjectSet()
    objset.AddObject(p)
    objset.AddObject(doc)

    // The creation info would not normally be added manually, but it has to be
    // since it's referenced by IRI (not by object) due to no deduplication
    // when serializing
    objset.AddObject(ci)


    file, err := os.Create("out.json")
    if err != nil {
        fmt.Println(err)
        os.Exit(1)
    }
    defer file.Close()

    encoder := json.NewEncoder(file)
    if err := objset.Encode(encoder); err != nil {
        fmt.Println(err)
        os.Exit(1)
    }

}
```

Note that serialization is still very incomplete in the go bindings, so this code does references by IRI instead of by object, but the places that affects are commented in the code

JPEWdev:
OK, I was motivated to do some cleanup here. The object references are now fixed, as well as few other things. Here is the new test code which works and passes SPDX 3 validation:
```go
package main

import (
    "time"
    "os"
    "fmt"
    "encoding/json"
    "model"
)

func main() {
    objset := v3_0_1.NewSHACLObjectSet()

    p := v3_0_1.MakePerson()
    p.ID().Set("http://spdx.org/spdxdocs/person-example")
    p.Name().Set("Joshua")
    objset.AddObject(p)

    ci := v3_0_1.MakeCreationInfo()
    ci.SpecVersion().Set("3.0.1")
    ci.CreatedBy().Append(v3_0_1.MakeObjectRef[v3_0_1.Agent](p))
    ci.Created().Set(time.Now().UTC())

    p.CreationInfo().Set(v3_0_1.MakeObjectRef(ci))

    email := v3_0_1.MakeExternalIdentifier()
    email.ExternalIdentifierType().Set(v3_0_1.ExternalIdentifierTypeEmail)
    email.Identifier().Set("noreply@noreply.com")

    p.ExternalIdentifier().Append(v3_0_1.MakeObjectRef(email))

    doc := v3_0_1.MakeSpdxDocument()
    doc.ID().Set("http://spdx.org/spdxdocs/doc-example")
    doc.CreationInfo().Set(v3_0_1.MakeObjectRef(ci))

    objset.AddObject(doc)

    file, err := os.Create("out.json")
    if err != nil {
        fmt.Println(err)
        os.Exit(1)
    }
    defer file.Close()

    encoder := json.NewEncoder(file)
    if err := objset.Encode(encoder); err != nil {
        fmt.Println(err)
        os.Exit(1)
    }

}
```

nishakm:
@JPEWdev So you can use `interface{}` instead of golang types:
```
// SHACL Property
type Property interface {
	Get()
	Set(val interface{}) error
}

type PropertyBase struct {
	value interface{}
	name  string
}

func (p *PropertyBase) Get() interface{} {
	return p.value
}

func (p *PropertyBase) Set(val interface{}) error {
	p.value = val
	return nil
}

func main() {
	prop1 := PropertyBase{}
	prop1.Set("hello")
	prop2 := PropertyBase{}
	prop2.Set(1)
}
```

But this is really annoying so I agree with you that we probably need to move to the latest Go version (1.23 at this time). We may need to host a new repo in spdx for this specifically (at least for now).

nishakm:
This looks right to me...
```
{
  "@context": "https://spdx.org/rdf/3.0.1/spdx-context.jsonld",
  "@graph": [
    {
      "creationInfo": {
        "@id": "_:0",
        "created": "2024-12-18T16:43:24Z",
        "createdBy": [
          "https://swinslow.net/spdx-examples/example1/hello-v3-specv3/SPDXRef-gnrtd0"
        ],
        "specVersion": "3.0.1",
        "type": "CreationInfo"
      },
      "externalIdentifier": [
        {
          "externalIdentifierType": "email",
          "identifier": "steve@swinslow.net",
          "type": "ExternalIdentifier"
        }
      ],
      "name": "Steve Winslow",
      "spdxId": "https://swinslow.net/spdx-examples/example1/hello-v3-specv3/SPDXRef-gnrtd0",
      "type": "Person"
    },
    {
      "creationInfo": "_:0",
      "spdxId": "https://swinslow.net/spdx-examples/example1/hello-v3-specv3/document0",
      "type": "SpdxDocument"
    }
  ]
}
```

nishakm:
@JPEWdev it looks like `SHACLObjectSet` doesn't have any methods to get individual objects? Is this something we want to implement?

nishakm:
My code right now:
```
  1 package main
  2 
  3 import (
  4         "encoding/json"
  5         "fmt"
  6         "os"
  7         spdx "spdx3/pkg/spdx"
  8         "time"
  9 )
 10 
 11 func main() {
 12         // Write doc
 13         //WriteToStdout()
 14 
 15         // read doc
 16         ReadFromStdin()
 17 
 18 }
 19 
 20 func WriteToStdout() {
 21         // not a document but a SHACLObjectSet
 22         objset := MakeDoc()
 23         // write to stdout
 24         encoder := json.NewEncoder(os.Stdout)
 25         if err := objset.Encode(encoder); err != nil {
 26                 fmt.Println(err)
 27                 os.Exit(1)
 28         }
 29 }
 30 
 31 func ReadFromStdin() {
 32         // read from stdin
 33         decoder := json.NewDecoder(os.Stdin)
 34         objset := spdx.NewSHACLObjectSet()
 35         if err := objset.Decode(decoder); err != nil {
 36                 fmt.Println(err)
 37                 os.Exit(1)
 38         }
 39         fmt.Println(objset)
 40 }
 41 
 42 func MakeDoc() spdx.SHACLObjectSet {
 43         // hmmm
 44         objset := spdx.NewSHACLObjectSet()
 45 
 46         p := spdx.MakePerson()
 47         // ought to be generated
 48         p.ID().Set("https://swinslow.net/spdx-examples/example1/hello-v3-specv3/SPDXRef-gnrtd0")
 49         p.Name().Set("Steve Winslow")
 50         // This is how you create a reference
 51         p_ref := spdx.MakeObjectRef[spdx.Agent](p)
 52         objset.AddObject(p)
```

Issue #268, opened 2025-10-28
ExternalDocumentRef.DocumentRefID does not perform required prefix processing because it is a string


KAWAHARA-souta:
ExternalDocumentRef.DocumentRefID is a string.

spdx/v2/v2_3/document.go (L.19~23):
```
type ExternalDocumentRef struct {
	// DocumentRefID is the ID string defined in the start of the
	// reference. It should _not_ contain the "DocumentRef-" part
	// of the mandatory ID string.
	DocumentRefID string `json:"externalDocumentId"`
```

Although the comment states "It should not contain the 'DocumentRef-' part", this processing is not performed. 
The type of this value should be common.DocElementID.Retry

KAWAHARA-souta:
Resolved in #269

Issue #267, opened 2025-10-28
Fix broken tagvalue example used in tests & Remove the tagvalue-specific prefix processing


KAWAHARA-souta:
The value of the ExternalDocumentRef field in the tagvalue example, which is also used in tests, is clearly broken. 

examples/sample-docs/tv/SPDXTagExample-v2.3.spdx (L.6):
```
ExternalDocumentRef: DocumentRef-DocumentRef-spdx-tool-1.2 http://spdx.org/spdxdocs/spdx-tools-v1.2-3F2504E0-4F89-41D3-9A0C-0305E82C3301 SHA1:d6a770ba38583ed4bb4525bd96e50461655d2759
```

You can clearly see that "DocumentRef-" appears twice.
This example is used in tests, and unfortunately, the tests pass.

Actually, the DocumentRefID of ExternalDocumentRef is currently a string, and while json and yaml do not perform prefix processing, only tagvalue has an implementation that performs tagvalue-specific prefix processing.

spdx/v2/v2_3/document.go (L.19~23):
```
type ExternalDocumentRef struct {
	// DocumentRefID is the ID string defined in the start of the
	// reference. It should _not_ contain the "DocumentRef-" part
	// of the mandatory ID string.
	DocumentRefID string `json:"externalDocumentId"`
```
(Note that although the comment states "It should not contain the 'DocumentRef-' part", no such processing is actually performed because it's a string. This will be addressed in a another issue. (#268))

First, to align the processing across formats, it is necessary to fix the ExternalDocumentRef in the example to the correct string and remove the tagvalue-specific prefix processing.

KAWAHARA-souta:
Resolved in #269

PR #269, opened 2025-10-28
Resolve two issues related to ExternalDocumentRef

KAWAHARA-souta:
ExternalDocumentRef.DocumentRefID does not perform required prefix processing because it is a string, although comments states like below.

spdx/v2/v2_3/document.go (L.19~23):
```
type ExternalDocumentRef struct {
	// DocumentRefID is the ID string defined in the start of the
	// reference. It should _not_ contain the "DocumentRef-" part
	// of the mandatory ID string.
	DocumentRefID string `json:"externalDocumentId"`
```

I've fixed it and also fixed related issue. (#267, #268)

### DIFF
--- a/spdx/v2/common/identifier.go
+++ b/spdx/v2/common/identifier.go
@@ -30,11 +30,7 @@ func (d *DocumentID) UnmarshalJSON(data []byte) error {
 	idStr := string(data)
 	idStr = strings.Trim(idStr, "\"")
 
-	e, err := trimDocumentIdPrefix(idStr)
-	if err != nil {
-		return err
-	}
-	*d = e
+	*d = trimDocumentIdPrefix(idStr)
 	return nil
 }
 
@@ -47,15 +43,9 @@ func prefixDocumentId(id DocumentID) string {
 	return val
 }
 
-// trimDocumentIdPrefix removes the DocumentRef- prefix from an document ID string or returns an error if it
-// does not start with DocumentRef-
-func trimDocumentIdPrefix(id string) (DocumentID, error) {
-	// handle DocumentRef-
-	if !strings.HasPrefix(id, documentRefPrefix) {
-		return "", fmt.Errorf("failed to parse SPDX identifier '%s'", id)
-	}
-	e := DocumentID(strings.TrimPrefix(id, documentRefPrefix))
-	return e, nil
+// trimDocumentIdPrefix removes the DocumentRef- prefix from an document ID string
+func trimDocumentIdPrefix(id string) DocumentID {
+	return DocumentID(strings.TrimPrefix(id, documentRefPrefix))
 }
 
 // ElementID represents the identifier string portion of an SPDX element
@@ -75,11 +65,7 @@ func (d *ElementID) UnmarshalJSON(data []byte) error {
 	idStr := string(data)
 	idStr = strings.Trim(idStr, "\"")
 
-	e, err := trimElementIdPrefix(idStr)
-	if err != nil {
-		return err
-	}
-	*d = e
+	*d = trimElementIdPrefix(idStr)
 	return nil
 }
 
@@ -92,15 +78,9 @@ func prefixElementId(id ElementID) string {
 	return val
 }
 
-// trimElementIdPrefix removes the SPDXRef- prefix from an element ID string or returns an error if it
-// does not start with SPDXRef-
-func trimElementIdPrefix(id string) (ElementID, error) {
-	// handle SPDXRef-
-	if !strings.HasPrefix(id, spdxRefPrefix) {
-		return "", fmt.Errorf("failed to parse SPDX identifier '%s'", id)
-	}
-	e := ElementID(strings.TrimPrefix(id, spdxRefPrefix))
-	return e, nil
+// trimElementIdPrefix removes the SPDXRef- prefix from an element ID string
+func trimElementIdPrefix(id string) ElementID {
+	return ElementID(strings.TrimPrefix(id, spdxRefPrefix))
 }
 
 // DocElementID represents an SPDX element identifier that could be defined
@@ -141,7 +121,7 @@ func (d DocElementID) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON takes a SPDX Identifier string parses it into a DocElementID struct.
 // This function is also used when unmarshalling YAML
-func (d *DocElementID) UnmarshalJSON(data []byte) (err error) {
+func (d *DocElementID) UnmarshalJSON(data []byte) error {
 	// SPDX identifier will simply be a string
 	idStr := string(data)
 	idStr = strings.Trim(idStr, "\"")
@@ -158,11 +138,7 @@ func (d *DocElementID) UnmarshalJSON(data []byte) (err error) {
 		// an SPDXRef can appear after a DocumentRef, separated by a colon
 		idFields = strings.SplitN(idStr, ":", 2)
 
-		d.DocumentRefID, err = trimDocumentIdPrefix(idFields[0])
-		if err != nil {
-			return err
-		}
-
+		d.DocumentRefID = trimDocumentIdPrefix(idFields[0])
 		if len(idFields) == 2 {
 			idStr = idFields[1]
 		} else {
@@ -170,8 +146,8 @@ func (d *DocElementID) UnmarshalJSON(data []byte) (err error) {
 		}
 	}
 
-	d.ElementRefID, err = trimElementIdPrefix(idStr)
-	return err
+	d.ElementRefID = trimElementIdPrefix(idStr)
+	return nil
 }
 
 // TODO: add equivalents for LicenseRef- identifiers
--- a/spdx/v2/common/identifier.go
+++ b/spdx/v2/common/identifier.go
@@ -14,6 +14,52 @@ const (
 	documentRefPrefix = "DocumentRef-"
 )
 
+// DocumentID represents the identifier string portion of an SPDX document
+// identifier. DocumentID should be used for reference to a SPDX document.
+// DocumentIDs should NOT contain the mandatory 'DocumentRef-' portion.
+type DocumentID string
+
+// MarshalJSON returns an DocumentRef- prefixed JSON string
+func (d DocumentID) MarshalJSON() ([]byte, error) {
+	return marshal.JSON(prefixDocumentId(d))
+}
+
+// UnmarshalJSON validates DocumentRef- prefixes and removes them when processing DocumentIDs
+func (d *DocumentID) UnmarshalJSON(data []byte) error {
+	// SPDX identifier will simply be a string
+	idStr := string(data)
+	idStr = strings.Trim(idStr, "\"")
+
+	e, err := trimDocumentIdPrefix(idStr)
+	if err != nil {
+		return err
+	}
+	*d = e
+	return nil
+}
+
+// prefixDocumentId adds the DocumentRef- prefix to an document ID if it does not have one
+func prefixDocumentId(id DocumentID) string {
+	val := string(id)
+	if !strings.HasPrefix(val, spdxRefPrefix) {
+		return documentRefPrefix + val
+	}
+	return val
+}
+
+// trimDocumentIdPrefix removes the DocumentRef- prefix from an document ID string or returns an error if it
+// does not start with DocumentRef-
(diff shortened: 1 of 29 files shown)


spdx/v2/v2_3/document.go
```diff
@@ -20,7 +20,7 @@ type ExternalDocumentRef struct {
 	// DocumentRefID is the ID string defined in the start of the
 	// reference. It should _not_ contain the "DocumentRef-" part
 	// of the mandatory ID string.
-	DocumentRefID string `json:"externalDocumentId"`
+	DocumentRefID common.DocElementID `json:"externalDocumentId"`
```
kzantow: `common.DocElementID` is the combination of 3 things: external document reference, element reference, and special ID. The `DocumentRefID` is supposed to be only an ID to reference the external document within this document, this should not include ElementID or special references, this is why it is a string. It is a little unclear if this should include `DocumentRef-`. If you look at [the JSON schema](https://github.com/spdx/spdx-spec/blob/support/2.3.1/schemas/spdx-schema.json#L102), it makes no reference to this except in the [license section](https://github.com/spdx/spdx-spec/blob/support/2.3.1/schemas/spdx-schema.json#L169), which indicates `DocumentRef-` is only used in an _element ID reference_ when you are referring to an external document using the `docrefIdString`, which I think is supposed to match the value here: `DocumentRef-[docrefIdString]:`.

kzantow (dismissed):
I do not think this change is correct. Adding an _element ID reference_ to the document ID specifier is not what we should do.

It's a little unclear to me whether the JSON external document reference should include `DocumentRef-` or not -- please note that the JSON format has some oddities that don't always match the TagValue and other formats, but I think it probably should based on [the examples in the spec](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#663-examples).

I think the correct change to make here is not to use the `common.DocElementID`, but instead create a new typed string of `type common.DocumentID string`, which has the behavior of prepending and removing the `DocumentRef-`, this field which is used in the External Reference objects, as well as the `common.DocElementID` for consistency. This would end up being less of a breaking change, too, since initializers using strings would continue to work.

KAWAHARA-souta:
Thanks @kzantow ,

I agree that creating a new typed string of `type common.DocumentID string`.
I will modify the PR accordingly. After that, Could you review it again once I complete the changes?


I also have a question regarding the creation of `type common.DocumentID string`.

Currently, DocElementID is structured as follows:
```
type DocElementID struct {
	DocumentRefID string
	ElementRefID  ElementID
	SpecialID     string
}
```

Would it be OK to change the type of `DocElementID.DocumentRefID` to the newly created `common.DocumentID` ?
Additionally, like my commit https://github.com/spdx/tools-golang/pull/269/commits/ea4f07f7888389885d5299c30b96dd3f27773db6 , would it be OK to make it possible to output a DocElementID as a simple Document ID even when it only has DocumentRefID ?

My main concern is that the type of items specified in Relationship is common.DocElementID:
```
type Relationship struct {
	// 11.1: Relationship
	// Cardinality: optional, one or more; one per Relationship
	//              one mandatory for SPDX Document with multiple packages
	// RefA and RefB are first and second item
	// Relationship is type from 11.1.1
	RefA         common.DocElementID `json:"spdxElementId"`
	RefB         common.DocElementID `json:"relatedSpdxElement"`
```

My main intention is to enable the creation of Relationships to Documents (including ExternalDocumentRefs)

KAWAHARA-souta:
Hi, @kzantow 
I've modified and force updated my PR. Could you review it?
I've tried to focus on the issues we need to fix now and make the PR simple.

convert/spdx_document_conversion_test.go
```diff
@@ -80,15 +80,15 @@ func Test_ConvertSPDXDocuments(t *testing.T) {
 				DocumentNamespace: "doc namespace",
 				ExternalDocumentReferences: []v2_1.ExternalDocumentRef{
 					{
-						DocumentRefID: "doc ref id 1",
+						DocumentRefID: common.DocumentID("doc ref id 1"),
```
kzantow: Are these casts needed? I don't think initializers need to have a cast. Same comment for all other initializers

spdx/v2/common/identifier.go
```diff
@@ -14,6 +14,52 @@ const (
...
+		return documentRefPrefix + val
+	}
+	return val
+}
+
+// trimDocumentIdPrefix removes the DocumentRef- prefix from an document ID string or returns an error if it
+// does not start with DocumentRef-
+func trimDocumentIdPrefix(id string) (DocumentID, error) {
+	// handle DocumentRef-
+	idFields := strings.SplitN(id, documentRefPrefix, 2)
```
kzantow: I think this could just. be `strings.TrimPrefix(id, documentRefPrefix)`

spdx/v2/v2_1/tagvalue/reader/parse_creation_info.go
```diff
@@ -103,7 +103,7 @@ func (parser *tvParser) parsePairFromCreationInfo(tag string, value string) erro
 
 // ===== Helper functions =====
 
-func extractExternalDocumentReference(value string) (string, string, string, string, error) {
+func extractExternalDocumentReference(value string) (common.DocumentID, string, string, string, error) {
```
kzantow: Do these functions even need to change? Could there just be a cast to string somewhere? (Same question for other similar tagvalue changes)

spdx/v2/common/identifier_test.go
```diff
@@ -39,7 +39,7 @@ func Test_DocElementIDEncoding(t *testing.T) {
 			value: DocElementID{
 				DocumentRefID: "a-doc",
 			},
-			err: true,
+			expected: "DocumentRef-a-doc",
```
kzantow: I think this should still be an error -- there is no ElementID specified

kzantow (commented):
This looks a lot closer, I noticed there is a test that was returning an error which no longer is, I think that case is actually still supposed to be an error, so that will need to be updated, unless I've misunderstood. Also, I think a lot of these changes could be undone -- initializers don't need the casts, do they? And I think the tagvalue changes could be simplified even more to just cast the value from the struct to/from strings rather than updating everything to the new type, since these are just serialization codes, right?

spdx/v2/common/identifier_test.go
KAWAHARA-souta: Hmm, as I mentioned in my previous comment, this was the one of what I wanted to implement..

> Additionally, like my commit https://github.com/spdx/tools-golang/commit/ea4f07f7888389885d5299c30b96dd3f27773db6 , would it be OK to make it possible to output a DocElementID as a simple Document ID even when it only has DocumentRefID ?
> My main concern is that the type of items specified in Relationship is common.DocElementID:
> 
> ```
> type Relationship struct {
> 	// 11.1: Relationship
> 	// Cardinality: optional, one or more; one per Relationship
> 	//              one mandatory for SPDX Document with multiple packages
> 	// RefA and RefB are first and second item
> 	// Relationship is type from 11.1.1
> 	RefA         common.DocElementID `json:"spdxElementId"`
> 	RefB         common.DocElementID `json:"relatedSpdxElement"`
> ```
> My main intention is to enable the creation of Relationships to Documents (including ExternalDocumentRefs)

For now, revert it back as you suggested.

convert/spdx_document_conversion_test.go
KAWAHARA-souta: Sorry, I got confused with the previous fix and included too many extra casts.
I've fixed it.

spdx/v2/common/identifier.go
KAWAHARA-souta: fixed it in https://github.com/spdx/tools-golang/pull/269/commits/d31a3d6e5cd1ee6d347b774a608a58c4ca8cfdd4
This implementation was brought over from trimElementIdPrefix(), so I've fixed it too.

spdx/v2/v2_1/tagvalue/reader/parse_creation_info.go
KAWAHARA-souta: fixed in https://github.com/spdx/tools-golang/pull/269/commits/6cdc30a1bb9a1e7d734e55aad1cbbbd6446af107

spdx/v2/common/identifier_test.go
kzantow: Sorry, I'm trying to understand what you are trying to do -- are you trying to refer to the document itself? It has an element ID of `SPDXRef-DOCUMENT`, I think there should always be an element referenced here.

spdx/v2/common/identifier.go
```diff
@@ -14,6 +14,50 @@ const (
...
+		return documentRefPrefix + val
+	}
+	return val
+}
+
+// trimDocumentIdPrefix removes the DocumentRef- prefix from an document ID string or returns an error if it
+// does not start with DocumentRef-
+func trimDocumentIdPrefix(id string) (DocumentID, error) {
+	// handle DocumentRef-
+	if !strings.HasPrefix(id, documentRefPrefix) {
```
kzantow: I think we should be lenient here, especially since it seems like we may have been outputting documents without `DocumentRef-`, just get rid of this check and the whole function should basically be `return DocumentID(strings.TrimPrefix(id, documentRefPRefix)), nil`.

kzantow (commented):
Looks great, just one small thing: we should be lenient in reading documents, I think

spdx/v2/common/identifier_test.go
kzantow: Maybe it would be worthwhile defaulting the ElementID to `SPDXRef-DOCUMENT` if the documentID is set but not the element ID? But I think it's very likely to be a programming error if the element ID is not specified...

spdx/v2/common/identifier.go
KAWAHARA-souta: OK, I've fixed, and also fixed trimElementIdPrefix() too.

spdx/v2/common/identifier_test.go
KAWAHARA-souta: I see, so references to the default document can be handled with SPDXRef-DOCUMENT...
I apologize. It seems I was severely mistaken.
When handling SPDX documents, you inevitably deal with relationships to the document itself, and I thought it would be useful to treat that as a DocumentID.
However, it's fine to just treat it as an ElementID like SPDXRef-DOCUMENT, isn't it...
I'm sorry, please disregard this.

spdx/v2/common/identifier.go
```diff
@@ -14,6 +14,44 @@ const (
...
+	return marshal.JSON(prefixDocumentId(d))
+}
+
+// UnmarshalJSON validates DocumentRef- prefixes and removes them when processing DocumentIDs
+func (d *DocumentID) UnmarshalJSON(data []byte) error {
+	// SPDX identifier will simply be a string
+	idStr := string(data)
+	idStr = strings.Trim(idStr, "\"")
+
+	e, err := trimDocumentIdPrefix(idStr)
```
kzantow: This will never return an error now, so could be simplified

spdx/v2/common/identifier.go
```diff
@@ -14,6 +14,44 @@ const (
...
+func prefixDocumentId(id DocumentID) string {
+	val := string(id)
+	if !strings.HasPrefix(val, spdxRefPrefix) {
+		return documentRefPrefix + val
+	}
+	return val
+}
+
+// trimDocumentIdPrefix removes the DocumentRef- prefix from an document ID string
+func trimDocumentIdPrefix(id string) (DocumentID, error) {
```
kzantow: I think this function doesn't even need to return an error now, sorry I should have realized this the first time I asked for changes 🤦

KAWAHARA-souta: You're right, I should have realized too..

kzantow (approved):
Thanks for the fix @KAWAHARA-souta!

spdx/v2/common/identifier_test.go
kzantow: To be clear: I don't really know what the right thing to do here is, but since we must refer to the document with `SPDXRef-DOCUMENT`, for example to specify the root relationships, I think it would implicitly have that ID in all cases, and if you were to refer to it with a document ref, it would be `<document-id>:SPDXRef-DOCUMENT`. I went ahead and merged this fix, as I think it corrects the main issue, we can address this separately, if this approach does not work for you, for some reason

PR #272, opened 2025-11-26
fix: Fix prefixDocumentId() to use correct prefix

KAWAHARA-souta:
prefixDocumentId() uses "spdxRefPrefix" in its prefix check.
But that was wrong. "documentRefPrefix" is the correct one.

And, Tests for DocumentID itself were not implemented. So, I've implemented it.

(This was my mistake in #269 , I'm very sorry... )


kzantow (approved):

kzantow:
Thanks for the fix @KAWAHARA-souta! :)

Issue #274, opened 2026-01-05
Incorrect successful parsing of invalid SPDX identifier fields


awyeth:
SPDX identifiers require the `SPDXRef-` prefix and DocumentRef identifiers require the `DocumentRef-` prefix.

See spec details [6.3](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#6.3), [7.2](https://spdx.github.io/spdx-spec/v2.3/package-information/#7.2), and [8.2](https://spdx.github.io/spdx-spec/v2.3/file-information/#8.2)

Currently, if the `SPDXRef-` prefix isn't present then the SBOM is parsed and the ID is accepted verbatim. Further, on being written out the SBOM will have the `SPDXRef-` prefix added.

Note, this also means that `SPDXRef-` is an accepted ID.

For example, the following SBOM:

```
{
  "SPDXID" : "SPDXRef-DOCUMENT",
  "spdxVersion" : "SPDX-2.3",
  "creationInfo" : {
    "created" : "2023-07-18T12:27:48Z",
    "creators" : [ "Person: Gary O'Neall", "Tool: spdx-maven-plugin" ],
    "licenseListVersion" : "3.21"
  },
  "name" : "tools-java",
  "dataLicense" : "CC0-1.0",
  "documentDescribes" : [ "SPDXRef-8" ],
  "documentNamespace" : "test",
  "packages" : [ {
    "SPDXID" : "7",
    "name" : "json-schema-validator"
  }],
  "relationships" : [ {
    "spdxElementId" : "7",
    "relationshipType" : "TEST_CASE_OF",
    "relatedSpdxElement" : "0"
  }]
}
```

is currently successfully parsed and when written has the following output:

```
{
  "spdxVersion":"SPDX-2.3",
  "dataLicense":"CC0-1.0",
  "SPDXID":"SPDXRef-DOCUMENT",
  "name":"tools-java",
  "documentNamespace":"test",
  "creationInfo":{
    "licenseListVersion":"3.21",
    "creators":["Person: Gary O'Neall","Tool: spdx-maven-plugin"],
    "created":"2023-07-18T12:27:48Z"
  },
  "packages":[{
    "name":"json-schema-validator",
    "SPDXID":"SPDXRef-7",
    "downloadLocation":"",
    "filesAnalyzed":true
  }],
  "relationships":[{
     "spdxElementId":"SPDXRef-7",
     "relatedSpdxElement":"SPDXRef-0",
[...]

PR #275, opened 2026-01-06
Fail parsing if the required prefix isn't present for IDs

awyeth:
If the `SPDXRef-` or `DocumentRef-` prefix isn't present in an ID, then fail parsing. In addition, if stripping the prefix leaves an empty ID then also fail.

This is in line with the spec:

https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#6.3
https://spdx.github.io/spdx-spec/v2.3/package-information/#7.2
https://spdx.github.io/spdx-spec/v2.3/file-information/#8.2

See https://github.com/spdx/tools-golang/issues/274 for more details about the issue being fixed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: '1.18'
+          go-version: '1.22'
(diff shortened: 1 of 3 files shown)


lumjjb (approved):

kzantow (approved):
LGTM

PR #276, opened 2026-01-16
chore: Update Release Notes v0.5.0 - v0.5.7

bact:
- Use release notes information from https://github.com/spdx/tools-golang/releases
- Use Markdown frontmatter (between `---` lines) for SPDX-License-Identifier metadata; this will remove Markdownlint warning of first line is not a title and the metadata will be rendered as a table on GitHub

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,36 +1,222 @@
+---
 SPDX-License-Identifier: CC-BY-4.0
+---
 
 # Release Notes for spdx/tools-golang
+
+## 0.6.0
+
+-rc2 released on: 2026-01-15
+-rc1 released on: 2026-01-09
+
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -1,36 +1,222 @@
...
+-rc2 released on: 2026-01-15
+-rc1 released on: 2026-01-09
+
+### What's Changed
+
+* feat: SPDX 3 by @kzantow in #247
+
+### Contributors
+
+* @kzantow
```
kzantow: These releases have not been released from `main`, so probably shouldn't show up in the release notes yet.

bact: You're right. 0.6.0 is removed for now.

PR #279, opened 2026-02-23
fix: handle ExternalDocumentRef tag in CreationInfo parser state

### LINKED ISSUE #195: SPDX parsing fails for ExternalDocumentRef in document.
The attached file fails to parse with the following error message (please remove .txt before processing)

```
Error: received unknown tag ExternalDocumentRef in CreationInfo section
```

ExternalDocumentRef is a valid tag with 1+ cardinality, so this appears to be a bug.

My guess is we are missing ExternalDocumentReference check in this enumeration of tags: https://github.com/spdx/tools-golang/blob/main/spdx/v2/v2_2/tagvalue/reader/parse_creation_info.go#L26-L101 

It got lost around this refactor: https://github.com/spdx/tools-golang/commit/460cf54ece7eba418c50407ee35544c5d63588a7#diff-dc282dc5935546282e5429e56eda94391c511ea039311c1d9db2f6501f730d4cL41 

[96b_aerocore2-hello_world-build.spdx.txt](https://github.com/spdx/tools-golang/files/10769488/96b_aerocore2-hello_world-build.spdx.txt)

dschmidt:
Fixes #195

The tag-value parser rejected valid SPDX documents where `ExternalDocumentRef` appeared after `Creator`/`Created` tags. The spec doesn't mandate field ordering, so the parser was too strict.

Adds `ExternalDocumentRef` as a recognized tag in `parsePairFromCreationInfo` for v2_1, v2_2, and v2_3 — using the same delegation pattern already used for `DocumentComment`.

### DIFF
--- a/spdx/v2/v2_1/tagvalue/reader/parse_creation_info.go
+++ b/spdx/v2/v2_1/tagvalue/reader/parse_creation_info.go
@@ -94,6 +94,9 @@ func (parser *tvParser) parsePairFromCreationInfo(tag string, value string) erro
 		return parser.parsePairForAnnotation(tag, value)
 	case "AnnotationComment":
 		return parser.parsePairForAnnotation(tag, value)
+	case "ExternalDocumentRef":
+		parser.st = psStart
+		return parser.parsePairFromStart(tag, value)
 	default:
 		return fmt.Errorf("received unknown tag %v in CreationInfo section", tag)
 	}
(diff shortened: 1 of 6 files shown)


kzantow (approved):
LGTM thanks, @dschmidt!

dschmidt:
you're welcome, thanks for reacting so quickly :)

PR #263, opened 2025-08-26
Fix GetAllFilePaths returning paths without slash prefix

OhioDschungel6:
This PR should fix https://github.com/spdx/tools-golang/issues/262

PR #281, opened 2026-03-27
Fix GetAllFilePaths slash prefix issue and resolve test failures from #263

### LINKED ISSUE #262: filesystem.GetAllFilePath violates its contract if the input path is "."
If you have "." as input path for filesystem.GetAllFilePath, the returned paths don't start with "/".

jijo-OO7:
Fixes #262

Normalize dirRoot using filepath.Abs before walking the directory to ensure
consistent path prefix handling for relative inputs such as ".".

Also updates tests by replacing t.Chdir with os.Chdir for compatibility with
older Go versions while preserving behavior.

Builds upon discussion in #263 and the earlier fix proposed by @OhioDschungel6.

### DIFF
--- a/utils/filesystem.go
+++ b/utils/filesystem.go
@@ -41,6 +41,10 @@ func GetAllFilePaths(dirRoot string, pathsIgnored []string) ([]string, error) {
 
 		path = filepath.ToSlash(path)
 		shortPath := strings.TrimPrefix(path, prefix)
+		if !strings.HasPrefix(shortPath, "/") {
+			//This happens if the dirRoot is "."
+			shortPath = "/" + shortPath
+		}
 
 		// don't include path if it should be ignored
 		if pathsIgnored != nil && ShouldIgnore(shortPath, pathsIgnored) {
--- a/utils/filesystem.go
+++ b/utils/filesystem.go
@@ -24,9 +24,14 @@ func GetAllFilePaths(dirRoot string, pathsIgnored []string) ([]string, error) {
 	// this is so that it can be appropriately modified by append
 	// in the sub-function.
 	paths := &[]string{}
+	absRoot, err := filepath.Abs(dirRoot)
+	if err != nil {
+		return nil, err
+	}
+	dirRoot = absRoot
 	prefix := strings.TrimSuffix(filepath.ToSlash(dirRoot), "/")
 
-	err := filepath.Walk(dirRoot, func(path string, fi os.FileInfo, err error) error {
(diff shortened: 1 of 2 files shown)


utils/filesystem.go
```diff
@@ -41,6 +41,10 @@ func GetAllFilePaths(dirRoot string, pathsIgnored []string) ([]string, error) {
 
 		path = filepath.ToSlash(path)
 		shortPath := strings.TrimPrefix(path, prefix)
+		if !strings.HasPrefix(shortPath, "/") {
```
kzantow: I think a better solution would be just to do a `dirRoot, err = filepath.Abs(dirRoot)` before the `filepath.Walk`. This should ensure it has a consistent absolute path prefix, regardless of what behavior a relative path might have; sorry I should have had a closer look before

jijo-OO7: > I think a better solution would be just to do a `dirRoot, err = filepath.Abs(dirRoot)` before the `filepath.Walk`. This should ensure it has a consistent absolute path prefix, regardless of what behavior a relative path might have; sorry I should have had a closer look before

Thanks, that makes sense  normalizing dirRoot upfront is a cleaner and more robust approach.

jijo-OO7:
@kzantow  I've updated the implementation to use filepath.Abs(dirRoot) before the walk and removed the conditional prefix handling. Let me know if this looks good.

jijo-OO7:
Hey @kzantow just a friendly ping would like your review on this when you have a moment 
Thanks!

jijo-OO7:
@kzantow  could you have a look at this when you have moment .
Thanks !

utils/filesystem.go
```diff
@@ -41,6 +46,7 @@ func GetAllFilePaths(dirRoot string, pathsIgnored []string) ([]string, error) {
 
 		path = filepath.ToSlash(path)
 		shortPath := strings.TrimPrefix(path, prefix)
+		shortPath = "/" + strings.TrimPrefix(shortPath, "/")
```
kzantow: Could we change this to only adding `"/"` when required? something like:
```
if !strings.HasPrefix(shortPath, "/") {
  shortPath = "/" + shortPath
}
```
? 

I think with change to an absolute path coupled with `filepath.ToSlash(dirRoot)`, this should always have a leading slash, so this is just creating new strings every time for no reason.

jijo-OO7: @kzantow  Thanks ! for the review i have updated the PR acordingly .

kzantow (approved):
LGTM, thanks @jijo-OO7

PR #283, opened 2026-04-13
fix: convert license expressions, fix conversion issues

kzantow:
This PR primarily corrects an issue where all licenses were converted from 2.3 to 3.0 as SimpleLicenses. Now, complex licenses are converted to the conjunctive/disjunctive license structures.

While working on this, I noticed tests were erroneously passing due to not comparing the `Elements` list. After discussion with implementors, it seems that Elements are where all elements should go which are not the direct subject of a collection and this should contain _all elements_ in collection's object graph. As an example, an SpdxDocument has Elements and RootElements, an SBOM would typically be the sole entry in the RootElements, the SBOM RootElements has a root package, say a package which represents the container. Relationships are added to the Elements list and this ends up being the only way to reference related packages and other related information. In addition to that, the Elements properties would contain the container package -- all the way up the object graph to the SpdxDocument.Element property. Now, any ElementCollection automatically has it's Elements collection flattened and deduplicated during serialization. These Element lists are also tested as part of the diffing process.

### DIFF
--- a/spdx/v3/v3_0/convert.go
+++ b/spdx/v3/v3_0/convert.go
@@ -61,6 +61,8 @@ func From_v2_3(doc v2_3.Document, d *Document) {
 	d.DataLicense = c.convert23licenseExpression(doc.DataLicense)
 
 	var converted ElementList
+
+	// other licenses may be referenced (with LicenseRef-... in expressions), so process these first to recreate the proper graph
 	for _, l := range doc.OtherLicenses {
 		converted = append(converted, c.convert23license(l))
 	}
@@ -836,7 +838,37 @@ func (c *documentConverter) convert23licenseExpression(licenseExpression string)
 	if err != nil {
 		c.logDropped(err)
 	}
-	return parsedLicense
+	return c.resolveLicenseRefs(parsedLicense)
+}
+
+// resolveLicenseRefs walks a parsed license expression and replaces the
+// placeholder CustomLicense references the parser produces for LicenseRef-*
+// identifiers with the resolved CustomLicense instances registered in
+// idMap by convert23license.
+func (c *documentConverter) resolveLicenseRefs(license AnyLicenseInfo) AnyLicenseInfo {
+	switch l := license.(type) {
+	case *CustomLicense:
+		if resolved, ok := c.idMap[l.ID].(*CustomLicense); ok {
+			return resolved
+		}
+	case *ConjunctiveLicenseSet:
+		for i, m := range l.Members {
+			l.Members[i] = c.resolveLicenseRefs(m)
+		}
+	case *DisjunctiveLicenseSet:
+		for i, m := range l.Members {
+			l.Members[i] = c.resolveLicenseRefs(m)
+		}
+	case *OrLaterOperator:
+		if resolved, ok := c.resolveLicenseRefs(l.SubjectLicense).(AnyLicense); ok {
+			l.SubjectLicense = resolved
+		}
+	case *WithAdditionOperator:
+		if resolved, ok := c.resolveLicenseRefs(l.SubjectExtendableLicense).(AnyExtendableLicense); ok {
+			l.SubjectExtendableLicense = resolved
+		}
+	}
+	return license
 }
 
 func (c *documentConverter) convert23externalDocumentRef(r v2_3.ExternalDocumentRef) AnyExternalMap {
--- a/spdx/v3/v3_0/convert.go
+++ b/spdx/v3/v3_0/convert.go
@@ -258,7 +258,6 @@ type documentConverter struct {
 	externalRefTypeMap        map[string]ExternalRefType
 	primaryPurposeMap         map[string]SoftwarePurpose
 	emailExtractor            *regexp.Regexp
-	postConvert               []func() error
 	conversionErrors          []error
 }
 
--- a/spdx/v3/v3_0/convert.go
+++ b/spdx/v3/v3_0/convert.go
@@ -243,8 +243,7 @@ func newDocumentConverter(d *Document) *documentConverter {
 }
 
 type documentConverter struct {
-	sbom *SBOM
-	//namespace                 string
+	sbom                      *SBOM
 	idMap                     map[string]any
 	creationInfo              AnyCreationInfo
 	relationshipMap           map[any][]AnyRelationship
@@ -846,8 +845,15 @@ func (c *documentConverter) convert23licenseExpression(licenseExpression string)
 // idMap by convert23license.
 func (c *documentConverter) resolveLicenseRefs(license AnyLicenseInfo) AnyLicenseInfo {
 	switch l := license.(type) {
+	case *ListedLicense:
+		if l.ID != "" {
+			if existing, _ := c.idMap[l.ID].(AnyListedLicense); existing != nil {
+				return existing
+			}
+			c.idMap[l.ID] = l
+		}
 	case *CustomLicense:
-		if resolved, ok := c.idMap[l.ID].(*CustomLicense); ok {
+		if resolved, ok := c.idMap[l.ID].(AnyCustomLicense); ok {
 			return resolved
 		}
 	case *ConjunctiveLicenseSet:
--- a/spdx/v3/v3_0/convert.go
+++ b/spdx/v3/v3_0/convert.go
@@ -58,7 +58,7 @@ func From_v2_3(doc v2_3.Document, d *Document) {
 	d.Comment = doc.DocumentComment
 	d.Imports = list[ExternalMapList](c.convert23externalDocumentRef, doc.ExternalDocumentReferences...)
 	d.Name = doc.DocumentName
-	d.DataLicense = c.convert23licenseString(doc.DataLicense)
+	d.DataLicense = c.convert23licenseExpression(doc.DataLicense)
 
 	var converted ElementList
 	for _, l := range doc.OtherLicenses {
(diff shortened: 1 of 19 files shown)


spdx/v3/v3_0/convert_test.go
```diff
@@ -132,16 +295,10 @@ func diffOpts() []cmp.Option {
 			cmpopts.IgnoreFields(t, "ID", "CreationInfo"),
 		)
 	}
-	for _, t := range []any{
-		SpdxDocument{},
-		SBOM{},
-		Bundle{},
-	} {
-		out = append(out,
-			cmpopts.IgnoreFields(t, "Elements"),
```
kzantow, line 141: This was causing a lot of stuff to be missed in the diff.

spdx/v3/v3_0/license_parser.go
```diff
@@ -0,0 +1,234 @@
...
+
+func isIdentChar(c byte) bool {
+	return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '-' || c == '.'
+}
+
+// makeLicense creates the appropriate license type based on the identifier prefix.
+// LicenseRef-* and DocumentRef-* identifiers produce CustomLicense, all others produce ListedLicense.
+func makeLicense(ident string) AnyLicense {
+	if strings.HasPrefix(ident, "LicenseRef-") || strings.HasPrefix(ident, "DocumentRef-") {
+		return &CustomLicense{ID: ident}
```
wagoodman: This function here is pure on it's own but the way it's used by callers might not be correct. If we create two custom license when converting based off of the string, but they have different associated metadata, then you might drop or override one (up in convert.go).

wagoodman: ok, take 2, I ran out of time when first investigating: 

When say... a `LicenseRef-Foo` appears both as an entry in `OtherLicenses` (with name/text/comment/seeAlso populated) and inside a license expression on a Package (e.g. `PackageLicenseDeclared: "MIT AND LicenseRef-Foo"`), we end up with two separate `CustomLicense` objects sharing the same ID. 

[`convert23License` builds the populated one and stashes it in `c.idMap`](https://github.com/kzantow-anchore/tools-golang/blob/8dc81deb5c0c3147864ded0070f3dc2491778413/spdx/v3/v3_0/convert.go#L815-L825), but `convert23licenseExpression` calls `ParseLicenseExpression`, which inside `makeLicense` [unconditionally does `&CustomLicense{ID: ident}`](https://github.com/kzantow-anchore/tools-golang/blob/8dc81deb5c0c3147864ded0070f3dc2491778413/spdx/v3/v3_0/license_parser.go#L221)... a bare instance with no name, no text, no seeAlso. That stripped instance is what ends up wired into the relationship graph, while the populated one is orphaned.

spdx/v3/v3_0/convert_v301_test.go
```diff
@@ -233,22 +264,36 @@ func v301package2() (*Package, ElementList) {
...
+	declaredLicense := &DisjunctiveLicenseSet{
+		Members: LicenseInfoList{
+			&ListedLicense{Name: "AGPL"},
+			&DisjunctiveLicenseSet{
+				Members: LicenseInfoList{
+					&WithAdditionOperator{
+						SubjectExtendableLicense: &ListedLicense{Name: "GPL-2.0-only"},
+						SubjectAddition:          &ListedLicenseException{Name: "Classpath-Exception"},
+					},
+					&ListedLicense{Name: "LicenceRef-CustomLicense1"},
```
wagoodman:
```suggestion
					&ListedLicense{Name: "LicenseRef-CustomLicense1"},
```

I think this is falling back to the `ListedLicense` and side stepping the validation here

spdx/v3/v3_0/license_parser.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func isIdentChar(c byte) bool {
+	return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '-' || c == '.'
+}
+
+// makeLicense creates the appropriate license type based on the identifier prefix.
+// LicenseRef-* and DocumentRef-* identifiers produce CustomLicense, all others produce ListedLicense.
+func makeLicense(ident string) AnyLicense {
+	if strings.HasPrefix(ident, "LicenseRef-") || strings.HasPrefix(ident, "DocumentRef-") {
```
wagoodman: Do we need to worry about `NOASSERTION` and`NONE` here? I think this is falling back to the ListedLicenses approach again

spdx/v3/v3_0/convert.go
```diff
@@ -666,28 +827,16 @@ func (c *documentConverter) convert23license(l *v2_3.OtherLicense) AnyLicenseInf
...
-			Name: licenseString,
-			Text: licenseString,
-		}
-	} else {
-		out = &ListedLicense{
-			Name: licenseString,
-		}
+	parsedLicense, err := ParseLicenseExpression(licenseExpression)
+	if err != nil {
+		c.logDropped(err)
```
wagoodman, line 832:
```suggestion
		c.logDropped(err)
		return nil
```

spdx/v3/v3_0/license_parser.go
```diff
@@ -0,0 +1,234 @@
...
+	if p.pos < len(p.input) && p.input[p.pos] == '+' {
+		p.pos++
+		return &OrLaterOperator{SubjectLicense: makeLicense(ident)}, nil
+	}
+
+	return makeLicense(ident), nil
+}
+
+func (p *licenseParser) skipWhitespace() {
+	for p.pos < len(p.input) && p.input[p.pos] == ' ' {
```
wagoodman: since this is saying "whitespace" I feel this should cover more cases (tab, newline, carriage return, etc)

spdx/v3/v3_0/license_parser.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	if end < len(p.input) && isIdentChar(p.input[end]) {
+		return false
+	}
+	return true
+}
+
+func (p *licenseParser) consumeKeyword(keyword string) {
+	p.skipWhitespace()
+	p.pos += len(keyword)
```
wagoodman: no request for change, just noting this is a little brittle; there is a convention that the caller calls `peekKeyword`, and any new caller added that doesn't do this might mess up `p.pos` here.

spdx/v3/v3_0/convert.go
```diff
@@ -247,53 +256,75 @@ type documentConverter struct {
 	externalRefTypeMap        map[string]ExternalRefType
 	primaryPurposeMap         map[string]SoftwarePurpose
 	emailExtractor            *regexp.Regexp
+	postConvert               []func() error
```
wagoodman: dead code?

spdx/v3/v3_0/license_parser.go
wagoodman: actually, there could be some additional test cases around this that could be added which would highlight this

spdx/v3/v3_0/license_parser_test.go
```diff
@@ -0,0 +1,274 @@
+package v3_0
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+func TestParseLicenseExpression(t *testing.T) {
```
wagoodman, line 11: Maybe a couple of extra cases to add (could be contrived though):
- `+` combined with conjunction: `MIT+ AND Apache-2.0+` and `(MIT OR GPL-2.0-only+) AND BSD-3-Clause`. The flattening + or-later interaction isn't directly exercised.
- WITH on a `LicenseRef`: `LicenseRef-MyLic WITH AdditionRef-MyExc`. Exercises `makeAddition`'s `AdditionRef-` / `LicenseRef-` / `DocumentRef-` prefix logic on the addition side.
- `DocumentRef-` with `+`: `DocumentRef-ext:LicenseRef-foo+`. Verifies the colon-aware ident scanner cooperates with the `+` suffix.
- WITH inside a paren-grouped OR: `MIT OR (GPL-2.0-only WITH Classpath-exception-2.0)`. Pins the documented precedence (WITH binds tighter than OR) under an explicit grouping.

wagoodman: actually maybe some additional negative cases here would be helpful:

- `MIT WITH (Classpath-exception-2.0)`: exception cannot be parenthesized.
- `MIT WITH GPL-2.0-only+`: `+` is not allowed on the exception side.
- `(MIT OR Apache-2.0) WITH Classpath-exception-2.0`: WITH applied to a license set; should error via the `AnyExtendableLicense` type assertion (erg, right?). Either way a dedicated case would help so that path doesn't regress silently.
- `MIT +` (space before `+`): should this be accepted or rejected? I don't really know, but we could pin down the behavior in a test with a comment.
- Empty parens `()` and `( )`.

spdx/v3/v3_0/convert_v23_test.go
```diff
@@ -154,14 +154,14 @@ func v23package2() *v2_3.Package {
...
-		FilesAnalyzed:             false,
+		FilesAnalyzed:             true,
 		IsFilesAnalyzedTagPresent: true,
 		PackageChecksums: []common.Checksum{
 			{Algorithm: common.SHA256, Value: "ffaabbcc11223344"},
 		},
 		PackageHomePage:         "https://tools.com/utility",
 		PackageLicenseConcluded: "Apache-2.0",
-		PackageLicenseDeclared:  "Apache-2.0",
+		PackageLicenseDeclared:  "AGPL OR (GPL-2.0-only with Classpath-Exception OR LicenceRef-CustomLicense1)",
```
wagoodman:
```suggestion
		PackageLicenseDeclared:  "AGPL OR (GPL-2.0-only with Classpath-Exception OR LicenseRef-CustomLicense1)",
```

spdx/v3/v3_0/license_parser_test.go
```diff
@@ -0,0 +1,274 @@
...
+					t.Fatalf("expected error for input %q, got nil", tt.expression)
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if d := cmp.Diff(tt.want, got, opts...); d != "" {
+				t.Errorf("ParseLicenseExpression(%q) mismatch (-want +got):\n%s", tt.expression, d)
+			}
```
wagoodman, line 530: nit: could use testify instead of t.Error/Fatall

spdx/v3/v3_0/convert.go
kzantow: I think an error here and partial license information could result, which is better than an error and no license information, sorta...

wagoodman (approved):

PR #286, opened 2026-06-03
spdxlib: allow NONE and NOASSERTION on the right side of relationships

### LINKED ISSUE #236: ValidateDocument() should allow NOASSERTION and NONE
According to the [Spec 11.1.1](https://spdx.github.io/spdx-spec/v2.2.2/relationships-between-SPDX-elements/#111-relationship-field)

> In cases where there are "known unknowns", the use of the keyword NOASSERTION can be used on the right hand side of a relationship to indicate that the author is not asserting whether there are other SPDX elements (package/file/snippet) that are connected by relationships or not. That is, there could be some, but the author is not asserting one way or another.

>Similarly, the use of the keyword NONE can be used to indicate that an SPDX element (package/file/snippet) has no other elements connected by some relationship to it.

Currently the the `ValidateDocument` func returns an error when RHS is one of the above.

---

Note: as described #231 ValidateDocument also fails when right hand side is an external reference. Both of these cases [we](https://sbom.observer) see in the wild.

arpitjain099:
## Summary

`ValidateDocument` checked `validElementIDs[relationship.RefA.ElementRefID]` (and `RefB`) unconditionally for every relationship. The tag-value parser stores `NONE` and `NOASSERTION` on the right side of a relationship as a `DocElementID` with `SpecialID` set and an empty `ElementRefID` (see `spdx/v2/common/identifier.go`). So a spec-valid document such as `Relationship: SPDXRef-Pkg CONTAINS NOASSERTION` parsed fine but then failed validation with the misleading `" used in relationship but no such package exists"` (the identifier was empty).

SPDX 2.3 section 11.1 permits `NONE` and `NOASSERTION` as relationship targets, so this was a false rejection of conformant documents.

## Change

In the relationship loop in `spdxlib/documents.go`, the existence check now only applies when the reference is an actual element ID (`SpecialID == ""`). A special value (`NONE`/`NOASSERTION`) is not an element reference, so it is skipped. The change is scoped to this issue only.

## Test

Added a table-driven test over `NONE` and `NOASSERTION` on a relationship's `RefB`. It fails on the current code and passes with the fix. The full `spdxlib` suite and `go test ./...` pass; `gofmt` and `go vet` are clean.

Fixes #236

### DIFF
--- a/spdxlib/documents.go
+++ b/spdxlib/documents.go
@@ -26,11 +26,13 @@ func ValidateDocument(doc *spdx.Document) error {
 	validElementIDs[common.MakeDocElementID("", "DOCUMENT").ElementRefID] = true
 
 	for _, relationship := range doc.Relationships {
-		if !validElementIDs[relationship.RefA.ElementRefID] {
+		// a special value such as NONE or NOASSERTION (SPDX 2.3 section 11.1)
+		// is not an element reference, so skip the existence check for it
+		if relationship.RefA.SpecialID == "" && !validElementIDs[relationship.RefA.ElementRefID] {
 			return fmt.Errorf("%s used in relationship but no such package exists", string(relationship.RefA.ElementRefID))
 		}
 
-		if !validElementIDs[relationship.RefB.ElementRefID] {
+		if relationship.RefB.SpecialID == "" && !validElementIDs[relationship.RefB.ElementRefID] {
 			return fmt.Errorf("%s used in relationship but no such package exists", string(relationship.RefB.ElementRefID))
 		}
 	}
--- a/spdxlib/documents.go
+++ b/spdxlib/documents.go
@@ -26,13 +26,23 @@ func ValidateDocument(doc *spdx.Document) error {
 	validElementIDs[common.MakeDocElementID("", "DOCUMENT").ElementRefID] = true
 
 	for _, relationship := range doc.Relationships {
-		// a special value such as NONE or NOASSERTION (SPDX 2.3 section 11.1)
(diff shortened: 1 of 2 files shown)


spdxlib/documents.go
```diff
@@ -26,11 +26,13 @@ func ValidateDocument(doc *spdx.Document) error {
 	validElementIDs[common.MakeDocElementID("", "DOCUMENT").ElementRefID] = true
 
 	for _, relationship := range doc.Relationships {
-		if !validElementIDs[relationship.RefA.ElementRefID] {
+		// a special value such as NONE or NOASSERTION (SPDX 2.3 section 11.1)
+		// is not an element reference, so skip the existence check for it
+		if relationship.RefA.SpecialID == "" && !validElementIDs[relationship.RefA.ElementRefID] {
```
kzantow: This change does not seem to be validating `NONE` and `NOASSERTION`. I also think these should only be allowed on _the right hand side_ (`RefB`) -- the spec notes this for `NOASSERTION` and although it says nothing about `NONE` in this regard, I think this is the expectation: at least one side needs a valid ID, allowing this as the first element would be unexpected.

arpitjain099:
Thanks for the review, good catch on both counts. I restricted NONE and NOASSERTION to the right side only, so a special value on RefA now returns an error, and RefB is validated to be exactly NONE or NOASSERTION rather than accepting any special value. Added test cases covering a special value on the left side and an unknown special value on the right. Rebuilt and the spdxlib suite plus go test ./... pass.

